PR #191, opened 2012-09-18
support for header i18n

ChaoTang:
The plugin supports i18n for the month and weekday names and the weekStart option, but not include header. So, I fix the problem as a suggestion to you.

The default is English ('en'),  

To add more languages, simply add a key to $.fn.datepicker.datepickerHeader, before calling .datepicker(). Example:

$.fn.datepicker.datespickerHeader['en'] = 'dates[this.language].months[month] +  " "+  year';

So, we can get the header text by call 'eval()'.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -457,8 +457,13 @@
 				}
 			}
 			return cls;
+
 		},
 
+                fillHead: function(month, year) {
+		    var text = dates[this.language].getHeader(month, year);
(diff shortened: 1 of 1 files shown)


vsn4ik:
@ChaoTang Use $.fn.datepicker.dates[yourLang].titleFormat.
Example: http://bootstrap-datepicker.readthedocs.org/en/master/i18n.html#i18n

@acrobat, @hebbet Close PR, please.

PR #248, opened 2012-11-07
Allow to specify the parent element (default: body)

wojcikstefan:
Under some circumstances, one does not want to append the datepicker to `<body>` (e.g. because of a complex structure of elements and corresponding `position` styles). This change allows to specify the parent element and doesn't break the positioning.


acrobat:
Closing this PR as #1227 replaces the placement changes

PR #307, opened 2013-01-08
1. In the hide function remove this.setValue(), 2. Show picker after … 

jsilvestri:
...itioning it for proper rendering with modals/multiple datepickers on a page.

PR #343, opened 2013-01-30
Remove setValue() from hide()

createbang:
When the calendar is blurred, it shouldn't set the value to the date that is selected.


acrobat:
@eternicode how do suggest to fix this issue? As in [PR #307](https://github.com/eternicode/bootstrap-datepicker/pull/307#discussion_r22339161) you said

> Removing only this line puts the event trigger under the above if clause, which is incorrect.

acrobat:
Closing this as the fork was deleted. Issue #1217 is a followup for this fix

PR #430, opened 2013-04-11
Fix for strange scrolling behavior in modal

### LINKED ISSUE #325: Datepicker in bootstrap-modal
Hi,

I'm using the bootstrap-modal extension from here https://github.com/jschr/bootstrap-modal/
When I attach a datepicker to an input within a modal the picker is positioned far away under the input. I've extended your 'place'-method. Maybe you need it:

```javascript
place : function() {
            if (this.isInline)
                return;
            var zIndex = parseInt(this.element.parents().filter(function() {
                return $(this).css('z-index') != 'auto';
            }).first().css('z-index')) + 10;
            var offset = this.component ? this.component.offset() : this.element.offset();
            var height = this.component ? this.component.outerHeight(true) : this.element.outerHeight(true);
            var isInModal = this.element.parents('.modal').length;
            var top = offset.top + height;
            if(isInModal) {
                top = top - 40;
            }
            this.picker.css({
                top : top,
                left : offset.left,
                zIndex : zIndex
            });
        },
```

timraymond:
Placing a datepicker as a component in a bootstrap modal would bind to
the body rather than the element of the modal. This would cause the
datepicker to scroll with the body while the modal remained in place.
This patch binds the datepicker dropdown to the element where it was
invoked and changes the computation of the offset to use position(),
since offset() was relative to the window. Fixes #325.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -237,7 +237,7 @@
 
 		show: function(e) {
 			if (!this.isInline)
-				this.picker.appendTo('body');
+				this.picker.appendTo(this.element);
 			this.picker.show();
 			this.height = this.component ? this.component.outerHeight() : this.element.outerHeight();
 			this.place();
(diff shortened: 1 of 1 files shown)


acrobat:
Closing this as the positioning should be fixed in PR #1227

Issue #165, opened 2012-08-19
Method to change the date manually


ghost:
I'd like to have the ability to change the date manually. I have a calendar system where the display is linked to the hash in the url, so if the user types #12-03-2012 at the end of the URL, then the display changes to the 12th April. However, I can't make the datepicker change to that date as well.

This would make this awesome! (If it doesn't have it already! :) )

PR #440, opened 2013-04-24
Allowing user to manually input date

someyoungideas:
This allows the date to be set from the user manually inputing the date
into the input box. This had been discussed previously in an issue and was somewhat answered on a side note. The change is actually from another fork, but had never been actually a pull request. Recently we had used this feature and it worked wonderfully for the purpose we had in mind.

Here is the previous issue that was involved with this.

[Issue #165](https://github.com/eternicode/bootstrap-datepicker/issues/165)

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -381,6 +381,7 @@
 
 			if(fromArgs) this.setValue();
 
+			var oldViewDate = this.viewDate;
 			if (this.date < this.startDate) {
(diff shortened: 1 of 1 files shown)


itsalongstory:
Any update?

itsalongstory:
2013 - 2021

itsalongstory:
related to #2629

cfarren:
I've just discovered this as well!

PR #483, opened 2013-05-31
Place support for RTL

dralshehri:
added calculation of right offset for RTL calendar place function


acrobat:
Closing this a PR #1192 fixed this issue

PR #515, opened 2013-06-18
datesDisable option

matiasalvarez87:
An array to specify dates to disable. It also adds the .disabled-date class in order to customise the UI.

Usage:

```javascript
var input = $('<input />').datepicker({
                    format: 'yyyy-mm-dd',
                    datesDisabled: ['2012-10-1', '2012-10-10', '2012-10-20']
                }),
```


acrobat:
Thanks @matiasalvarez87!

PR #546, opened 2013-07-18
Right align calendar on right window edge conflict

Munter:
Right align calendar instead of aligning left and nudging when the left aligned calendar would cross the right window edge.

This looks prettier because the right calendar edge aligns with the component and the arrow correctly points to the component.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -421,12 +421,18 @@
 			// auto x orientation is best-placement: if it crosses a window
 			// edge, fudge it sideways
 			else {
-				// Default to left
-				this.picker.addClass('datepicker-orient-left');
-				if (offset.left < 0)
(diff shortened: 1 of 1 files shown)


acrobat:
@Munter can you please provide a jsfiddle illustrating the problem? You can use [this jsfiddle](http://jsfiddle.net/g02L8z8y/) as base

Munter:
Here you go: http://jsfiddle.net/g02L8z8y/1/

When the input is close to the right window edge, the edge detection sees that, but instead of reversing the alignment to avoid the comflict, it nudges the datepicker until it's inside the window. However the arrow now points at nothing.

datepicker

This patch inverts the alignment in the same situation, causing the date picker to align with the right edge, and the arrow to correctly point at the input

acrobat:
Thanks @Munter!

Issue #399, opened 2013-03-20
setDate() does not update the datepicker view


omidkrad:
#### REPRO STEPS
1. Using the UI, select datepicker then select a date.
2. Use datepicker API to set date to a different date using setDate function.
3. In the UI open datepicker again.
#### EXPECTED

See date on the calendar as specified in step 2.
#### ACTUAL

The date is not updated and shows the date selected in step 1.

PR #554, opened 2013-07-24
Fix setDate not updating the UI when settingDate on different Month

Dudemullet:
When using setDate for a month currently not visible, the highlight box will aparently move, but not the visible month. for now I have tracked the problem down to the `this.viewDate` not being updated before calling the `this.fill()` method.

This PR fixes issue #399

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -800,6 +800,8 @@
 				this.date = new Date(date);
 			if (!which || which  == 'view')
 				this.viewDate = new Date(date);
+			else
+				this.viewDate = this.date;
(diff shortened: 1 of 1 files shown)


alon1980:
@eternicode  I know its been a while now but we came across the same issue with version 1.5.1 - When using the API to manually change the date (using setDate or Update) and the month is also changed, the UI displays the same month and upon exiting the datepicker UI the date refreshes and get back to the original month (day stays the same).
Also tried latest version and same results. Will be happy to get any help or direction. I assumed it was fixed but it looks like it not so what am I missing here?
Thanks

PR #582, opened 2013-08-05
Swallow ESC-key on hide

inukshuk:
When closing the picker because ESC was pressed, stop propagating the keydown event, because the datepicker has already 'consumed' the key event. If the event is allowed to bubble up enclosing modal windows or dialogs might be closed as well.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1331,8 +1331,10 @@
 
 		keydown: function(e){
 			if (!this.picker.is(':visible')){
-				if (e.keyCode === 40 || e.keyCode === 27) // allow down to re-show picker
+				if (e.keyCode === 40 || e.keyCode === 27) { // allow down to re-show picker
 					this.show();
+					e.stopPropagation();
+        }
 				return;
 			}
 			var dateChanged = false,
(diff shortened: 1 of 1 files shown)


js/bootstrap-datepicker.js
```diff
@@ -895,6 +895,7 @@
 			if (this.picker.is(':not(:visible)')){
 				if (e.keyCode == 27) // allow escape to hide and re-show picker
```
acrobat: I think you should add `{ .. } ` for the if statement as there are now 2 statements for this statement

acrobat:
@inukshuk Can you rebase the PR against the current 1.4 branch?

inukshuk:
I rebased with your 1.4 branch, let me know if something doesn't work!

acrobat:
Did you push that to your git repo? As github still states there is a conflict

inukshuk:
Yes, [it looks good to me](https://github.com/inukshuk/bootstrap-datepicker/commits/patch-1/js/bootstrap-datepicker.js) — does the [file itself](https://github.com/inukshuk/bootstrap-datepicker/blob/patch-1/js/bootstrap-datepicker.js) look OK to you?

If there's still a problem, how about I just create a new PR — this is only a trivial change anyway.

inukshuk:
Rebased this again.

hebbet:
LGTM, merged. 

Thank you :+1:

PR #642, opened 2013-09-26
Qualify events by namespace "bsDatepicker"

ZekeLu:
Otherwise, it will be at the risk to pollute events from other controls

This [fiddle](http://jsfiddle.net/zekelu/dfbfe/) illustrates how event namespace works

### DIFF
--- a/docs/events.rst
+++ b/docs/events.rst
@@ -5,6 +5,7 @@ Datepicker triggers a number of events in certain circumstances.  All events hav
 
     $('.datepicker').datepicker()
         .on(picker_event, function(e){
+            # all the events are qualified by the namespace `bsDatepicker`. `picker_event` looks something like this: `show.bsDatepicker`
             # `e` here contains the extra attributes
         });
 
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -295,6 +295,7 @@
 
 			this.element.trigger({
 				type: event,
+				namespace: 'bsDatepicker',
 				date: local_date,
 				format: $.proxy(function(altformat){
 					var format = altformat || this.o.format;


rodsouto:
thanks @ZekeLu , I was having problems with bootstrap modals too :+1:

skoub:
is there any plan to include this kind of patch to the trunk? On my side, to solve the problem, i added the bootstrap namespace style "show.bs.datepicker" and "hide.bs.datepicker" and it solved the problem.

```
show: function ()
{
    if (this.element.attr('readonly'))
        return;
    if (!this.isInline)
        this.picker.appendTo(this.o.container);
    this.place();
    this.picker.show();
    this._attachSecondaryEvents();
    this._trigger('show.bs.datepicker');
    if ((window.navigator.msMaxTouchPoints || 'ontouchstart' in document) && this.o.disableTouchKeyboard)
    {
        $(this.element).blur();
    }
    return this;
},

    hide: function ()
    {
        if (this.isInline)
            return this;
        if (!this.picker.is(':visible'))
            return this;
        this.focusDate = null;
        this.picker.hide().detach();
        this._detachSecondaryEvents();
        this.viewMode = this.o.startView;
        this.showMode();

        if (
            this.o.forceParse &&
            (
                this.isInput && this.element.val() ||
                this.hasInput && this.element.find('input').val()
            )
        )
            this.setValue();
        this._trigger('hide.bs.datepicker');
        return this;
    },
```

acrobat:
@skoub Yes we want to fix this long lasting issue, but changing the events would mean a BC break. So we are figuring out if we are going that in a 1.X release (which is agianst the "rules" of semver) or this should wait for a 2.0

skoub:
@acrobat you should wait for the 2.0. But i think you should not wait too long to release a 2.0 since that the datepicker is not fully fonctionnal with the boostrap 3 and people (like me) will look elsewhere to find a datepicker that will work perfectly with BS3.

acrobat:
Yes indeed that true! 1.4 should be out by the end of the month and then we should have atleast a few bugfix releases and maybe also a 1.5 but nothing that should hold us back to already start development on a 2.0 version.

I will discus this with the team and so we can start picking goals for that 2.0 release. Goals we need to develop or get help from the community in developing these features/changes

skoub:
great! maybe you could add a section on the the description page to explain this bug and how we can apply the correction ourself until the v2.0?

acrobat:
Good idea @skoub, do you maybe want to send a pr for that? Otherwise I will add it to docs before the 1.4 release!

skoub:
@acrobat, i prefer to let you decide how to explain it in the doc because i don't know where you want to go with the events namespace convention.

Ideally, i would put something that have good chance to appear in the next version. Would it be "show.bs.datepicker" or maybe "show**n**.bs.datepicker" since that the picker is already visible when the event is fired.

jwt02180:
I'd just like to add that I came across this issue with the event conflict between this and bootstraps modal. It was unexpected since there's no documentation of this being a possibility, and I only found out it by looking through the issues here.

It'd be much appreciated if something was added to the docs about this until it's fixed in a future version.

whaus7:
Had the same issue. That nameSpace fix worked like a charm for me. Will I need to update anything with my datepickers or no?

andrews05:
@acrobat I'm not sure that fixing this necessarily means a BC break. Existing code that is currently listening for just 'show' will still pick up any 'show' event regardless of namespace.

LastExile16:
I can't believe this issue existed since 2014 and not solved by datepicker team :/

acrobat:
@LastExile16 this is a foss project so we all can only spend the free time we have and the time we have left after other personal priorities.

I've personally been busy in my personal life and had some vacation in the last months. I'm going to try to spend some more time on the project in the coming months and have the last minor release out (1.7). After that release we can start working on 2.0 and have the namespace issue resolved.

vanduynslagerp:
The namespace should be added [here](https://github.com/uxsolutions/bootstrap-datepicker/blob/master/js/bootstrap-datepicker.js#L1524) as well.

sambenne:
2023 and this is still not fixed

TomePale:
Any chance for adding namespace fix soon ?

PR #680, opened 2013-10-31
Disable viewmode switch

ehtb:
I needed to disable the viewmode switch for a client, perhaps useful to others as well.


lenardpalko:
Why was this closed after all ? 
For my current project I have a use-case for this as the client wants to restrict selection of date just for one month so he doesn't want to be able to navigate through other months.

Thanks!

Issue #464, opened 2013-05-13
DatePicker on modal not displaying


olson2334:
Hello,

I am using your datepicker with a bootstrap modal and I cannot get the calendar to display.  I looked at issue #325 and applied the 2 lines that were changed.   Prior to applying the lines nothing would happen in the modal for those elements i want to have datepicker displayed.  After I applied the 2 line changes I am guessing the datepicker must be displaying off the modal some place as the modal applies both vertical and horizontal scrolls.  

I did download the latest version but still have the same issue.  Anyone have a fix for this by chance?  

Thanks.

prior to 2 applying the 2 lines in issue 325:
![image](https://f.cloud.github.com/assets/4417091/495653/6e6629b2-bbcd-11e2-99e6-4afcde0d45e0.png)

after applying and trying to active datepicker:
![image](https://f.cloud.github.com/assets/4417091/495658/89761212-bbcd-11e2-99a2-a8ce13aa564b.png)

webhacking:
i got same issue. thx : #https://github.com/eternicode/bootstrap-datepicker/pull/710

pkundan24:
for model use z-index:999

PR #710, opened 2013-11-26
z-index calculation should not stop at first element

fadomire:
i guess that when calculating proper z-index for the datepicker, we should find the max z-index value of parents() and not stop at first

related to #533 #170 #683 #678 #656 #665 #634 #621 #609 #464


Armstrongya:
@Offirmo Thanks for your solution, it works for me.

PR #722, opened 2013-12-03
Chainable methods

tomalec:
Hi again,

It's duplicate of pretty old #244, but hopefully this time PR is more suitable for [contribution](https://github.com/eternicode/bootstrap-datepicker/blob/master/CONTRIBUTING.md) ;-)

So again:

I get an idea (, and need ;) ) to chain datepicker methods. E.g.

``` javascript
var myDatepicker = $('.selector').data('datepicker')
                                    .setDate(defaultDate)
                                    .show()
```

### DIFF
--- a/tests/suites/methods.js
+++ b/tests/suites/methods.js
@@ -10,69 +10,114 @@ module('Methods', {
     }
 });
 
-// test('remove', function(){
+test('remove', function(){
+    var returnedObject = this.dp.remove();
+    // ...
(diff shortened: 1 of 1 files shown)


ragulka:
This seems to break the default jQuery chainability, for example:

```javascript
$('#myfield').datepicker('setDate', new Date()).show();
// does not work: Uncaught TypeError: undefined is not a function
```

Or am I missing something?

PR #754, opened 2013-12-19
Using {todayBtn: true} trigger changeDate

javarr:
when I use the "today" button with the setting {todayBtn: true}. This should not trigger the event changeDate, because the date in the input field does not change. It changes only with {todayBtn: "linked"}. In _setDate check if (which !== 'view')


acrobat:
Thanks @javarr!

PR #800, opened 2014-01-28
If the input field has the disabled attribute the datepicker won't show

Fed03:
In both situations:
```html
<div class="input-group date" id="from" data-date-format="dd/mm/yyyy">
	<input class="form-control" size="16" type="text" name="start_date" id="start_date" required disabled>
	<div class="input-group-addon">
		<span class="glyphicon glyphicon-calendar"></span>
	</div>
</div>
```

```html
<input class="form-control" size="16" type="text" name="lol" id="lol" required disabled>
```

adding the disabled attribute on the input field make the datepicker not shows himself.

I've also added a new method to enable the field
```js
$('#id').datepicker('enable');
```

thnx for your attention

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -93,8 +93,9 @@
 		this.element = $(element);
 		this.isInline = false;
 		this.isInput = this.element.is('input');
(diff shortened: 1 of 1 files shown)


acrobat:
Fixed in #1731

PR #802, opened 2014-01-31
Option to prevent date picker to show when input field receives focus

jarannilsen:
Added a new option showOnFocus which can be set to false to prevent the date picker from showing when the input field associated with it receives focus. 

There are 5 failing tests, but these are also failing in master, so I have not done anything with them at the moment. 

3 new tests have been added to verify the new functionality.

Feedback appreciated!


.gitignore
```diff
@@ -1,3 +1,7 @@
 instrumented/
 tests/coverage.html
 docs/_build
+
+/.idea/
+
+*.iml
```
acrobat: Please remove these directories, you should add them to you local global .gitignore

acrobat:
Please fix the note I added and rebase against the current master. Then the PR should be ready for a final review!

jarannilsen:
Sorry about that, will have a look at it soon as I am able.

acrobat:
That's great @jarannilsen! Please also take a look at the travis-ci build as there are some jshint issues ;)

tests/suites/options.js
```diff
@@ -641,3 +641,54 @@ test('Multidate Separator', function(){
...
+            .appendTo('#qunit-fixture')
+            .val('2014-01-01')
+            .datepicker({
+            }),
+        dp = input.data('datepicker'),
+        picker = dp.picker;
+
+    input.focus();
+
+    ok(picker.is(":visible"), "Datepicker is hidden");
```
acrobat: Shouldn't this message state "Datepicker is visible"?

tests/suites/options.js
```diff
@@ -641,3 +641,54 @@ test('Multidate Separator', function(){
...
+            .val('2014-01-01')
+            .datepicker({
+                showOnFocus: true
+            }),
+        dp = input.data('datepicker'),
+        picker = dp.picker;
+
+    input.focus();
+
+    ok(picker.is(":visible"), "Datepicker is hidden");
```
acrobat: Shouldn't this message state "Datepicker is visible"?

tests/suites/options.js
```diff
@@ -641,3 +641,54 @@ test('Multidate Separator', function(){
...
+            .val('2014-01-01')
+            .datepicker({
+                showOnFocus: false
+            }),
+        dp = input.data('datepicker'),
+        picker = dp.picker;
+
+    input.focus();
+
+    ok(picker.is(":hidden"), "Datepicker is visible");
```
acrobat: Shouldn't this message state "Datepicker is hidden"?

jarannilsen: Hi, i believe the message is the message displayed if the test fails - i.e.
it should indicate the incorrect state of the value you are asserting.
However, I am not very familiar with this testing framework, and based my
tests on the previous tests in the suite, so please correct me if I am
wrong :)

acrobat: I'm also getting to know the qunit framework, but as far as I can see the message you can give (optional) to the ok method, is a message to show when the test is done (instead of the result). See [qunit::ok docs](http://api.qunitjs.com/ok/), they describe the message as "A short description of the assertion"

So i guess the messages have to change

jarannilsen:
Updated test messages.

acrobat:
Thanks @jarannilsen!

Issue #808, opened 2014-02-03
All modes visible at once


tgirardi:
About a year ago I took Eyecon's code and added a feature that allows to see all three modes at the same time:

![screen shot 2014-02-03 at 14 23 58](https://f.cloud.github.com/assets/2540464/2067314/7ec8057a-8cf9-11e3-91d2-513d31e60a2e.png)

I've been using that feature in a web app i've been developing since and it is very useful in order to allow a quicker selection of dates and a fast way of knowing the currently selected date.

Would you consider adding this feature to this project? If so, I could make a Pull Request, probably this week.

Also, because I based my code in Eyecon's version, I'm worried that there are new things to consider when writing this feature using this project's master branch as the base for the code, this time. So any suggestion would be much appreciated.

PR #815, opened 2014-02-11
Allow an horizontal layout (all pickers visible)

tgirardi:
Implement an option called "horizontal" which makes all three pickers
(years, months, days) visibles at once and laid out in an horizontal
fashion.

Note: as described in issue #808


tgirardi:
Any chance of revamping this Pull Request? I can help. Just tell me what you need me to improve

PR #822, opened 2014-02-14
added option 'appendWidgetTo' and fixed calculation of display position

pkracht:
Attaching the datepicker container always to the 'body' does not work as expected, if the input element is positioned in a scrollable parent container using 'overflow: auto'. The datepicker stays on his absolute position. I fixed the positioning by calculating the left and top values relative to the 'appendWidgetTo' container (default 'body'). Next fix will be the detection of scrollLeft, too (sorry, no time for now).

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -63,7 +63,7 @@
 				this.push.apply(this, new_array);
 			},
 			clear: function(){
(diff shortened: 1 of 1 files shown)


acrobat:
Created a new pull request for these changes in combination with the new container option, see #1227

PR #826, opened 2014-02-18
Prevent ENTER propagation on form

fabdouglas:
On "ENTER" keydown on picker, prevent the closest form to be submitted too.
Updated PR of #674

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1200,6 +1200,8 @@
 					this.fill();
 					if (this.picker.is(':visible')){
 						e.preventDefault();
+						e.stopPropagation && e.stopPropagation();
+						e.cancelBubble = true; // IE6,7,8 support
 						if (this.o.autoclose)
 							this.hide();
 					}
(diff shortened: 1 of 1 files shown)


js/bootstrap-datepicker.js
```diff
@@ -1200,6 +1200,8 @@
 					this.fill();
 					if (this.picker.is(':visible')){
 						e.preventDefault();
+						e.stopPropagation && e.stopPropagation();
+						e.cancelBubble = true; // IE6,7,8 support
```
acrobat: Yes please replace the check of `e.stopPropagation` and the `e.cancelBubble = true` with a single  `e.stopPropagation()` call

acrobat:
I came by this piece of info, so the combination of preventDefault + stopPropagation should fix this issue! 

> event.preventDefault(): Prevents the default action of the event from triggering. Does not stop the event > propagation to parent DOM elements.
> event.stopPropagation(): Prevents the default action of the event from propagating to parent DOM  elements.

@fabdouglas please fix the note added above and then this PR should be ready to merge :)

acrobat:
Thanks @fabdouglas!

PR #834, opened 2014-02-22
Allow passing templates for heading, content and footer of Datepicker as options. 

smolnar:
Should we allow passing entire custom templates for heading, content or footer or just parts of templates like previous and next buttons?


acrobat:
Closing this PR as #1700 got merged

PR #848, opened 2014-03-07
fixes window.resize bug that freezes ie7

corrego:
On ie7, the call to place() would generate a new window.resize event which would in turn call place() again and so on, generating an endless loop and freezing the browser. This change checks for an actual window resizing before calling place(), fixing that problem

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -63,7 +63,7 @@
 				this.push.apply(this, new_array);
 			},
 			clear: function(){
-				this.length = 0;
+				this.splice(0);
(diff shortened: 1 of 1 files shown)


acrobat:
Can you provide a jsfiddle example showing this problem? (you can use this [base jsfiddle](http://jsfiddle.net/g02L8z8y/) As I can't reproduce the problem on a local vm

acrobat:
Closing this PR for now as I can't reproduce it. If this is still an issue on the latest master, please reopen reopen the pull request and provide a jsfiddle example!

PR #853, opened 2014-03-13
Update bower.json

jstoup111:
Match the dependency to bootstrap that is in the docs.  Since this supports bootstrap 2 and 3 we shouldn't have a dependency requiring 3 to be used.


acrobat:
Thanks @jstoup111!

PR #872, opened 2014-03-29
Adding parseDate option

ccubley:
I added a parseDate option that takes a function. If this option is set, the function will be called prior to calling DPGlobal.parseDate. If it returns a value, that value will be used as the parsed value. If it returns undefined, DPGlobal.parseDate will be called to parse the value. This option will allow completely custom date parsing logic.

My use case for this is that our app needs to take strings like '0227' and infer the date 2/27/2014 based on other context we have available in the app. This doesn't fit into the realm of custom formatting nicely so I need an option to completely override the date parsing logic when needed.

All existing test pass. I also added two new tests to cover this area. First, pass parseDate function that returns undefined. Second, pass parseDate function that actually parses the date. Both new tests pass.

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -151,6 +151,15 @@ A space-separated string consisting of one or two of "left" or "right", "top" or
 
 .. _startdate:
 
+
+parseDate
+---------
+
+Function(date, format, language).  Default: undefined
+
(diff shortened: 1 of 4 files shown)


vsn4ik:
Use http://bootstrap-datepicker.readthedocs.io/en/stable/options.html#format.
Available after merge #1582.

@acrobat, @hebbet I believe that it is necessary to close this PR.

PR #883, opened 2014-04-04
Add toggleActive option

ximi:
This option prevents the date from becoming unset when clicking on the currently active date.
The current default is set to match the previous behavior.
The option is ignored when multidate is active (as there would be no easy way to unselect a date, once selected).

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -56,7 +56,7 @@ toggleActive
 
 Boolean. Default: true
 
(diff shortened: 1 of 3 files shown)


acrobat:
@ximi can you rebase your pull request against master, as in #843 we remove this behavior but an option like this looks like a more complete solution!

acrobat:
Thanks  @ximi!

PR #893, opened 2014-04-11
Option default view date

michaeljb:
When the calendar is initially opened, go to a day other than "today".


acrobat:
Thanks @michaeljb!

PR #908, opened 2014-04-23
Add new option and fix posotion problem with body > 100%

KuJIbka:
Hello.

Datepicker check scrollTop, but not scrollLeft, I add this.

Also, when datepicker use in popup window, which has display:fixed, position of datepicker on scroll is broke, I add new option 'hideOnScroll'.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -346,7 +346,6 @@
 					}, this)
 				}]
 			);
-
 			this._secondaryEvents = [
 				[this.picker, {
 					click: $.proxy(this.click, this)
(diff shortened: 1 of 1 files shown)


acrobat:
@KuJIbka can you provide a jsfiddle example for what you are trying to fix? (You can use this [base jsfiddle](http://jsfiddle.net/g02L8z8y/) to start from)

KuJIbka:
Hello.

Yes, here link <http://jsfiddle.net/g02L8z8y/16/>.

2015-01-03 17:52 GMT+03:00 Jeroen Thora <notifications@github.com>:

vsn4ik:
@KuJIbka, already fixed in master.
Before (your example): http://jsfiddle.net/g02L8z8y/16/
Now: http://jsbin.com/mibuzebovi/1/edit?html,css,js,output

PR #909, opened 2014-04-23
Add timepicker

rchavik:
This is based on bufferapp's fork applied on top of master.
It has been rewritten to display seconds instead of meridiem.

Usage example:

``` js
    $('#NodeCreated').datepicker({
        showTime: true,
        autoclose: false, // when showTime is true, autoclose is ignored
        todayBtn: true,
        todayHighlight: true,
        clearBtn: true,
        format: 'yyyy-mm-dd H:i:s'
    });
```

Here's how it looks:

screenshot from 2014-04-23 19 52 47


Hoverbear:
@acrobat I see this is no longer on the 1.5.0 milestone. Do you need anything else to help land this?

russplaysguitar:
This would be a nice feature to have

acrobat:
Yes indeed, I just think this PR needs some more fine tuining. For example the option to disabled/hide seconds

And we should check if @guypaskar's problem is solved

nhouse:
@russplaysguitar I found this fork and am currently testing it out: http://www.malot.fr/bootstrap-datetimepicker/

canfiax:
can we have this pulled into main repo please?

canfiax:
I tried pulling in latest commit using:

    "bootstrap-datepicker": "git@github.com:rchavik/bootstrap-datepicker.git#5d82d35f54c25b8cebda0c0f7493a39a80f1c4e8",

However the datetimepicker does not have a time field, even though i copied your example in first post. When I click at a date, it populates the input field like this:

    2015-12-18 H:i:s

What am I doing wrong?

canfiax:
It seems the latest commit / version does not contain any dist files other than:

    dist/css/bootstrap-datepicker.standalone.css

The JS is missing from dist.

canfiax:
Also when changing the dates, the "time" is lost.

I am using commit:

5d82d35f54c25b8cebda0c0f7493a39a80f1c4e8

acrobat:
Time support is something we would like to add in 2.0+

deepakmahakale:
I am still in need of the timepicker.
I just have to use another library because of timepicker requirement.

Otherwise, I have all the options `toDisplay` `toValue` which I need in this library

PR #922, opened 2014-05-07

alx249:
fix bug on enter date from keyboard, then select it from dialog


acrobat:
@alx249 can you provide a jsfiddle example for what you are trying to fix? (You can use this [base jsfiddle](http://jsfiddle.net/g02L8z8y/) to start from)

hebbet:
closing old tickets, with no newer reply.

PR #924, opened 2014-05-08
Add locales to main endpoints

scyros:
To use locales inside bower-install grunt task is necessary to include it into bower.json.


acrobat:
Closing this PR as the main option should only include the file that are necessary to run this plugin. Locale files are optional!

> main string or array: The primary acting files necessary to use your package.

PR #950, opened 2014-05-28
fixed issue with up and left arrow navigation if we have a start date

CubaSAN:
There are the problem that up end left arrow doesn't work when datepicker has a start date defined and the start date equals focused date by default. Fix is described in subject.


acrobat:
Closing this PR as #1165 fixed this issue

PR #955, opened 2014-05-29
Fix an issue with Twitter.BootstrapRTL

modeeb:
When using bootstrap-rtl.css (http://www.nuget.org/packages/Twitter.BootstrapRTL/) which replaces left's with right's, the datepicker has two issues:
• dropdown is expanded to page width (as in the pic below), so I appended "col-md-2" class to template
• after fixing width, setting .left property does not change position, so .right property needs to be set
datepicker


acrobat:
Closing this PR as this issue was fixed in #1192

PR #961, opened 2014-06-03
Replaced named entities with decimal entities

SlimerDude:
Named entities (such as &nbsp;) are not valid in XHTML pages but decimal entities (such as &#160;) are, so this simple enhancement lets the datepicker work in pages served as XHTML.

Without it, upon initialisation, errors are thrown:
 - Firefox 29: SyntaxError: An invalid or illegal string was specified
 - Chrome 35: Uncaught SyntaxError: Failed to set the 'innerHTML' property on 'Element': The provided markup is invalid XML, and therefore cannot be inserted into an XML document


acrobat:
Ok fair enough! Thanks @SlimerDude!

PR #968, opened 2014-06-04
Manage disabled input

fabdouglas:
Prevent the datepicker to be displayed for disabled or readonly input.
Note, that may cause issue for users that want both disabled input and active component.
In this case, a new boolean option should specify the behavior.


acrobat:
Fixed in #1731

PR #985, opened 2014-06-18
Added an option for the format of the monthly view title: 'monthViewTitle'

feeva:
Also, symbol for Korean is ko, while kr is for the country. I removed and added Korean translation files as so.


vsn4ik:
@feeva Use `$.fn.datepicker.dates[yourLang].titleFormat`.
Example: http://bootstrap-datepicker.readthedocs.org/en/master/i18n.html#i18n

@acrobat, @hebbet Close PR, please.

PR #1001, opened 2014-07-07

RichardFrontwise:
The click handler conflicts if a user has a broad click handler attached to eg the body.
e.stopPropagation(); prevents the event from bubbling to any user defined click handlers.


acrobat:
Thanks @RichardFrontwise!

PR #1007, opened 2014-07-12
Added bootstrap arrow icon on each side of the calendar.

zozoh94:
I replaced your charcarters for the left and the right with Arrows Icons from Bootstrap glyphicons.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1598,9 +1598,9 @@
 		},
 		headTemplate: '<thead>'+
 							'<tr>'+
-								'<th class="prev">&laquo;</th>'+
+								'<th class="prev"><span class="glyphicon glyphicon-arrow-left"></span></th>'+
 								'<th colspan="5" class="datepicker-switch"></th>'+
-								'<th class="next">&raquo;</th>'+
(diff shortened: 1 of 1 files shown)


acrobat:
I think we should combine this with a icon-lib option, where you can define which icon-lib you use. As many people don't use the default glyphicon's from twbs but use fontawesome. And we should have a fallback so the current characters used for prev/next are used by default

acrobat:
If we can support multiple icon libs + a fallback, this PR fixes #1007

Azaret:
What about instead of supporting multiple icon libs, add some `templates` option, like Twitter typeahead ?
```js
$('#sandbox-container input').datepicker({
  templates: {
    left-arrow: '<i class="fa fa-arrow-left"></i>',
    right-arrow: '<i class="fa fa-right-left"></i>'
  }
});
```

Since it is first a datepicker for Bootstrap, it make sense that the default is glyphicon. Then we can add some snippet for FA and other libs in the sandbox.

acrobat:
That would indeed be a solution to this problem!

Azaret:
I'll work on that next then.

Azaret:
Even without the `template` thing, replacing the arrow entities with glyphicons break the `click` event.
So this PR can be closed, I'm gonna make the changes tomorrow.

PR #1016, opened 2014-07-27
add amd version

infacq:
learn from https://github.com/umdjs/umd


amessinger:
+1

I would be in favor of this particular version : https://github.com/umdjs/umd/blob/master/jqueryPluginCommonjs.js

It would allow for both global, amd and commonjs module definitions.

amessinger:
Also could be a drop in replacement for the `/js/bootstra-datepicker.js` file.

acrobat:
I like the aproach of #1196 a bit more, but I will look into this for 1.5.0

acrobat:
Closing this PR for now as I think that #1196's fix is a better solution then 2 separate files

Issue #736, opened 2013-12-10
noConflict with datepicker range mode


PSDCoder:
Hi. When I have tried use datepicker in range mode with "noConflict" I have got this error:
```js
Uncaught TypeError: Property 'datepicker' of object [object Object] is not a function
```
My code is pretty simple:
```html
<body>
    <div id="booking-room-date-range" class="input-daterange">
        <input type="text" name="start" />
        <span class="add-on">to</span>
        <input type="text" name="end" />
    </div>
    <script src="../jquery/jquery.min.js"></script>
    <script src="js/bootstrap-datepicker.js"></script>
    <script>
        (function($, undefined) {
            $(function(){
                $.fn.bootstrapDP = $.fn.datepicker.noConflict();
                $("#booking-room-date-range").bootstrapDP();
            });
        })(window.jQuery);
    </script>
</body>
```

acrobat:
Fixed in #1117

Issue #1112, opened 2014-10-05
if using noConflict the date range fails because it's calling directly on JQuery


SafwanHak:
the following code fails when using noConflict along with DateRangePicker.  

$(this.inputs)
			.datepicker(options) <<<<<<<<<<<<<<
			.bind('changeDate', $.proxy(this.dateUpdated, this));

acrobat:
Closing this as a duplicate of #736

PR #1117, opened 2014-10-14
Fix for #1112, assigning plugin to a local variable

mfunkie:
Assigning the plugin to a local variable so that we can call it without invoking it as a jQuery plugin.  Before this, noConflict was not really noConflict as there was still a reliance on .datepicker()


acrobat:
This should also fix #736

acrobat:
Thanks @mfunkie!

PR #1136, opened 2014-10-27
Do not show datepicker for readonly inputs

kix:
Currently, calling `$(el).datepicker()` on a readonly input (`<input type="text" readonly/>`) enables users to modify its content. This patch disables the datepicker popup from showing when an `<input/>` is readonly.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -408,6 +408,8 @@
 		},
 
 		show: function(){
+			if (this.element.attr('readonly'))
+				return;
 			if (!this.isInline)
 				this.picker.appendTo('body');
 			this.picker.show();


MuyBien:
This code will not work if the datepicker class is append to a div (as in this example with the component : ). 
I came up with this code. I think it needs to be changed before merged but it works : 
```
    var dateInput = this.isInput ? this.element : this.element.find("input");
    if (dateInput.attr('readonly'))
        return;
```
I didn't modify the this.element because it will not work in the case you click the component element (this.element = [])

natsu90:
It should be disabled on `disabled` attribute, instead of `readonly`. Readonly is useful to disallow user to edit datepicker input directly; `02/26/201` is treated as a valid date.

PR #1165, opened 2014-11-22
Fixed bug. Keybord left and up do not work after clicking right/down. In...

wishmaster80:
Keybord left and up do not work after clicking right/down. In case when startDate = +1 and date was not set. I think that method dateWithinRange get wrong arument. newDate which is allways default date +-1. It should be newViewate, because arrows navigate on view.
I found bug on IE11. And after fix test on lates version of Firefoz and Chrome.

Best regards
Adam Pryka
wishmaster80


acrobat:
Thanks @wishmaster80!

Issue #1080, opened 2014-09-08
changeDate event doesn't return "date" if it is the same date.


rey-wright:
So if you do

````javascript
$('.element').datepicker().on("changeDate", function(e){
    console.log(e.date);
});
````
If you choose a different day other than today, it will return the date. If you choose the same date, it will return with 

````javascript
undefined
````

http://jsfiddle.net/tuG6C/333/ <--- click on the input field, click on dates, check out the console and see how it reacts. It will return the date unless it's the same date.

This doesn't seem logical. If it returns the date for all other dates, it should return for same day. I haven't tested other events but I know it happens on changeDate for sure.

acrobat:
Can anyone verify if this is still an issue? I can't reproduce it in chrome/ff with http://jsfiddle.net/vh1oknf7/5/

kaspiCZ:
Tested it out too. On the same env I had the bug on. Works fine for me now.

felipe-canaviri:
The fix provided by @sularome is working pretty well. Thanks !

PR #1190, opened 2014-12-15
fix for #1080

sularome:
Fix for #1080


acrobat:
@sularome can you add a unit test to test this fix?

acrobat:
Awaiting feedback in #1080 if this is still an issue

PR #1219, opened 2015-01-05
Added docs for datesDisabled #515

acrobat:
Added docs for the new option introduced in #515

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -72,6 +72,14 @@ Days of the week that should be disabled. Values are 0 (Sunday) to 6 (Saturday).
 .. figure:: _static/screenshots/option_daysofweekdisabled.png
     :align: center
 
+.. _datesdisabled:
+
+datesDisabled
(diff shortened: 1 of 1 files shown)


korun:
Should we add lib version in doc like this?

acrobat:
What do you mean with lib version @korun?

korun:
@acrobat, some kind of notice that this functionality is only available for version 1.4.0 or higher.

acrobat:
I think that's not a big issue, but something we could do in the future. Currently the docs are available for each version and the latest docs in master, so if someone uses the datepicker they should check the latest docs for the version they are using

korun:
I think this will be useful, for example http://api.jquery.com/ do this, and it's really helpful. :)

acrobat:
That's something to keep in mind for future features! If you want you can send a PR to fix the current docs with versions for the features added in 1.4.0!

PR #1236, opened 2015-01-14
Update bower.json dependencies

GabrielGil:
The official bootstrap package can be avoided for those who are using the official bootstrap-sass source. No need to download the LESS version.


acrobat:
You are right about the dependency, thanks @GabrielGil!

PR #1242, opened 2015-01-15

CheapSk8:
Updated z-index functionality in "place" function to check for explicit z-index set through stylesheet. Current func. retrieves z-index based on DOM stack. In some instances you might want to specify the z-index through a CSS class instead. Updated func. will first reset picker z-index and check for z-index set via stylesheet. If not-specified, DOM stack is used.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -593,7 +593,7 @@
 				appendOffset = $(this.o.container).offset();
 			// get zIndex from CSS or parent stack - allows explicit set in stylesheet
 			this.picker.css('z-index','');	// reset zIndex
-			var zIndex = (this.picker.css('z-index') != 'auto' ? this.picker.css('z-index') : (function(element) {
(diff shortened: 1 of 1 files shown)


acrobat:
@CheapSk8 can you rebase your PR against the current master?

acrobat:
ping @CheapSk8

CheapSk8:
Sorry, I have been away on a project. I just merged the latest master down and am going to see if I can fire up my test app again and test this.

CheapSk8:
I updated and resubmitted as #1419

rosskevin:
I have a situation where a sibling element (not parent) is absolutely positioned.  In this rare case, hardcoding the z-index with `!important` works i.e.

```css
.datepicker-dropdown {
  z-index: 1036 !important;
}
```

Since `!important` works, it could be argued that this pull may not be needed.  On the other hand, many people don't know about `!important`, so this pull request is helpful.   Either way, I wanted to note my current workaround for the case of overlapped absolutely positioned components that are not in the same stack.

PR #1252, opened 2015-01-26
Update french translation

lherry:


acrobat:
The first capital letter was removed not so long ago in #1239. He provided a wikipedia link why the months should be in lowercase. Dunno if you're french and agree on that?

lherry:
Effectively, I just go to the wiki page given by @janv8000 that explains it, but this is true when the date is enrolled in a sentence unlike english where day or month names always should start with uppercase (even in a sentence)...

But we are in another case : day or month names are alone in the sentence and, even if they don't have to start with an uppercase, like in all other cases, we should have the first letter of first word of the sentence in uppercase...

We just have to note that this case is not usual and may only be found on the date pickers-like, that's probably why it is not explained on the Wiki page...

janv8000:
I don't believe that the argument that the dates are part of a sentence is valid here.

Eg. 
image

"janvier 2015" is not really a sentence as it does not contain a subject, verb, ... So the rule for uppercasing the first letter does not really apply.

lherry:
I ask the question directly to The French Academy... Let will see !! :wink:

lherry:
I've spoke about it to many people who're agree with me: all write's startings should start with an uppercase even if the word usually don't have to start with one...

Anyway, I have no answer from French Academy at the moment... I still waiting and hope they'll answer me... (I've asked from the dedicated form that they offer)

In fact, we're both right: 
- I was speaking about the calendar picker view where days or months are alone, so, to me, it should starts with an uppercase
- You was speaking about full date value (i.e. : "Lundi 12 mars") : if I'm right, only first word should start with an uppercase (day) but not the others (month), so my new french definition is wrong...

This problem doesn't exist with english date because all words starts with an uppercase...

So, to add an uppercase to sentence's first letter's first word, when I will have confirmation, the solution would be a new small functionality rather than my update...

Are you agree with that?

janv8000:
Could you post some screenshots showing your intended behaviour of the datepicker (with different casings in different situations)?

When looking at other applications for inspiration you can see they also use lowercase for the names in french:

eg. [Globalize.js](https://github.com/jquery/globalize#cldr) uses the [Unicode CLDR](http://www.unicode.org/cldr/charts/26/by_type/date_&_time.gregorian.html#2bd0ab9fba1a49d8)
Chrome native datepicker: 
image

lherry:
I just received an answer from the french academy: a pickup calendar with months and/or day names seems to be considered as enumerations, so it doesn't need a first uppercase letter which seems consistent.

However, like I thought, a french full date should start with an uppercase letter (ie. "Lundi 9 février 2015") even if day or month names should not start with an uppercase letter.

So in our case, it should be applicable on full date result field format. So if the widget result a "ddd dd mmm yyyy" format, it should result a first uppercase letter. @janv8000, your first screenshot example ("Date de clôture") should start with an uppercase too.

Even if this behavior could be implemented on the widget, I don't know if it's worth because it means that the host application have to be able to parse the date with this particular case too... Maybe too much stuff for little need...

I wait a few to cancel my pull request, according your opinion !

acrobat:
I think we should keep the lower case names for now, but thanks for your contribution @lherry!

Azaret:
> a pickup calendar with months and/or day names seems to be considered as enumerations, so it doesn't need a first uppercase letter which seems consistent.

Like the wiki page linked at 786e26b4fe7dd9eb9a4dc795cab755c73a76f8aa states (and unlike the commit comment), there is no official rules for this. The only official rule about dates if that we should not capitalize month name in a full date.

But on the datepicker there is no full date, it's just month and day names. And in this case it's either fine to write it with upper or lower case.

For the dots, like I said in #1352, per design there is none in bs-dp. English too should use dots for abbreviation, yet there is none, by design. It seems ridiculous to have some language respecting the dots and not others, either we should put dots everywhere, either don't have them in French.

I really think those changes are rubbish, it's fine to write either uppercase in this case, and having a coherent design inside bs-dp seems more important to me.

lherry:
The full date appears after you've selected a date from the the datepicker in the result field (depending the given format)...

Azaret:
Yeah, I forgot to write that part my bad. For the result I agree to lowercase, but not for the picker.

lherry:
Your purpose seems to unify the translation because of no standards, right ? Even if I personaly prefer the uppercase on months and days, why should it be chosen rather than the lowercase ? Same question about dots ?

To conclude, if we unify the translations, what to choose ? Dot or not ? Uppercase or not ?

As I previously said, to simplify answer parsing, all in lower case seems easier to code (fortunately, it's already done like that, so no work to do).

Anyway, the simplest way to unify it seems to conform with solution the most used today in translations !

Azaret:
My purpose is to unify because of design. And yeah we can permit ourselves this because there is no standards or mighty rule about this (At least in French).

So it's all a matter of taste I guess, and the look and feel aimed by bs-dp. And I'm talking about the picker on itself, not the output. I would say that for the output it might need to more complexity than just dots or not and upper or lower case. But I would be a dick to easy coding if I said that...

Same thing for abbreviation by the way, I find ugly the 3/4/5 notation for months, you have month on 3 letters, others on 4, some with dots, others without. It's not sexy to be honest, but maybe I care too much about how things look. Anyway if bs-dp is going to keep this abbreviation rules it should be "Avr." not "Avril".

So yeah, my point is design for the picker, "standards" for the output. Isn't Bootstrap all about design ?

janv8000:
@azaret Could you post a jsfiddle with your proposed changes?

Issue #1199, opened 2014-12-18
Bower issue with 1.3.1


niemyjski:
I just updated to 1.3.1 and now my build is failing with: 

bower bootstrap-datepicker#~1.3.1     invalid-meta bootstrap-datepicker is missing "ignore" entry in bower.json
[11:17:05][Step 2/4] bower bootstrap-datepicker#~1.3.1         resolved git://github.com/eternicode/bootstrap-datepicker.git#1.3.1
[11:17:06][Step 2/4]

PR #1267, opened 2015-02-15
Refactoring less and update dist filenames.

vsn4ik:
- <s>Bower not supported `ignore` option</s>
- <s>`user-select` without prefix not exists (http://caniuse.com/#feat=user-select-none)</s>
- Fixed Bootstrap 2 `.add-on` cursor and `.add-on i` position
- Fixed Bootstrap 3 `.input-group-addon` cursor
- Fixed "dist" filenames in `bower.json` and `composer.json`
- Fixed some `csslint` errors
- Rename `docs/REAME.md` to `docs/README.md`
- Add `clearDates` to docs
- Move `.jscsrc` to `js/.jscsrc`


bower.json
```diff
@@ -1,11 +1,14 @@
...
+    "dist/css/bootstrap-datepicker.css",
+    "dist/css/bootstrap-datepicker3.css",
+    "js/bootstrap-datepicker.js"
+  ],
   "license": "Apache-2.0",
   "dependencies": {
     "jquery" : ">=1.7.1",
     "bootstrap" : ">=2.0.4 <4.0"
-  },
-  "ignore": {}
```
acrobat: I've added this empty ignore tag because of #1199, so this should be preserved to avoid error's when running bower install

vsn4ik:
Thanks! Сorrected `ignore` to array.

less/datepicker3.less
```diff
@@ -50,26 +50,21 @@
...
-	}
-	table{
+	table {
 		margin: 0;
 		-webkit-touch-callout: none;
 		-webkit-user-select: none;
 		-khtml-user-select: none;
 		-moz-user-select: none;
 		-ms-user-select: none;
-		user-select: none;
```
hebbet: Why did you remove that?

vsn4ik: This property without vendor prefix not exist (not supported by someone).

vsn4ik: https://developer.mozilla.org/en-US/docs/Web/CSS/user-select

acrobat:
Looks good at first, will check whole PR tonight and then it should be ready to merge!

less/datepicker3.less
hebbet: ok, bootstrap also adds user-input without prefixes

https://github.com/twbs/bootstrap/blob/master/less/mixins/vendor-prefixes.less#L222

acrobat: Maybe we should keep the property as it does no harm to the code. Csslint also throws a warning for it when the `user-select`property is not set.

```
WARNING: Missing standard property 'user-select' to go along with '-moz-user-select'. When using a vendor-prefixed property, make sure to include the standard one. (vendor-prefix) Browsers: All
```

so my idea is to keep the property

vsn4ik: Done.

acrobat:
Thanks @vsn4ik!

PR #1280, opened 2015-03-01
Added option to not show datepicker on readonly field

### LINKED ISSUE #1231: 1.3.1 - Won't popup if field is ReadOnly
It worked previously. When I removed ReadOnly, it worked, but the calendar was not correctly positioned.

acrobat:
This option fixes #1231. The default behavior is restored and displaying of the datepicker can be disabled for readonly inputs if a users wants it that way.

Not sure if the option name is a good name

### DIFF
--- a/tests/suites/options.js
+++ b/tests/suites/options.js
@@ -904,3 +904,32 @@ test('Default View Date', function(){
 
     equal(picker.find('.datepicker-days thead .datepicker-switch').text(), 'May 1977');
 });
+
+//datepicker-dropdown
+
+test('Enable on readonly options (default)', function(){
+    var input = $('<input readonly="readonly" />')
+            .appendTo('#qunit-fixture')
+            .datepicker({format: "dd-mm-yyyy"}),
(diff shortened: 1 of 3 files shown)


tests/suites/options.js
```diff
@@ -904,3 +904,34 @@ test('Default View Date', function(){
...
+test('Enable on readonly options (false)', function(){
+    var input = $('<input readonly="readonly" />')
+            .appendTo('#qunit-fixture')
+            .datepicker({
+                format: "dd-mm-yyyy",
+                enableOnReadonly: false
+            }),
+        dp = input.data('datepicker'),
+        picker = dp.picker,
+        target;
```
korun: Why we need `target` variable?

acrobat: Hmm I think it was copied from another test. I will remove it!

acrobat:
@korun fixed the unused variable!

korun:
burns-excellent

Issue #820, opened 2014-02-13
Scrolling on mobile device closes datepicker


berniegp:
Hello,

When opening a datepicker in a mobile device, the month's grid sometimes cannot be displayed completely (i.e. it's too tall). Scrolling the page to see the rest of the month's grid unfortunately closes the datepicker. This makes it awkward to use this module on mobiles devices. In my tests, changing the event that closes the datepicker from touchstart to touchend fixed the issue.

Here's the diff (sorry, about inline diff, but it's too small to bother forking) :

```diff
bootstrap-datepicker/js/bootstrap-datepicker.js
@@ -355,7 +355,7 @@
 					resize: $.proxy(this.place, this)
 				}],
 				[$(document), {
-					'mousedown touchstart': $.proxy(function(e){
+					'mousedown touchend': $.proxy(function(e){
 						// Clicked outside the datepicker, hide it
 						if (!(
 							this.element.is(e.target) ||
```

acrobat:
Fixed in #1284

PR #1284, opened 2015-03-03
Bug #820 correction

chandon:
On mobile, when scrolling, the popup was closed.
Simply don't use touchstart/touchend : mousedown will be trigger on touchend IF there is no scroll.
When there is a scroll event between touchstart and touchend, mouse isn't triggered.


acrobat:
Looks good @chandon, can you please squash you commits

acrobat:
Thanks @chandon. Merged in 30c2f1d

PR #1285, opened 2015-03-03
Include "main" property in package.json

olanb7:
This will enable the datepicker to be included via `require('bootstrap-datepicker')` when using browserify/webpack.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "datepicker",
     "twitter-bootstrap"
   ],
+  "main": "./dist/js/bootstrap-datepicker.js",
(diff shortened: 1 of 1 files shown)


panzerdp:
It would be great to have this merged :) Thanks.

acrobat:
Yes this is something to merge soon! I'm not sure if main should be `"main": "./dist/js/bootstrap-datepicker.js",` or `"main": "dist/js/bootstrap-datepicker.js". Or maybe that doesn't matter?

panzerdp:
@acrobat I think both options will work.

acrobat:
Ok thanks for the feedback @panzerdp! Thanks @olanb7 for the PR!

PR #1296, opened 2015-03-11
Use $.on() instead $.bind().

vsn4ik:
"As of jQuery 1.7, the .on() method is the preferred method for attaching event handlers to a document." (http://api.jquery.com/bind/)


acrobat:
Thanks @vsn4ik!

Issue #778, opened 2014-01-10
Inconsistent Copy / Paste Behavior


austenc:
When you copy/paste a date (like '09/04/1948') into a datepicker field with `Ctrl+V` and hit `Tab` the datepicker works as expected, updating on paste / keeping the date you pasted once you hit tab. 

However, If you (may have to refresh prior to this) paste that same date, 09/04/1948 into the field using **Right Click -> Paste** and then press `Tab` it will change the date you pasted to the default date.

Win7 x64, Google Chrome

[Feel free to confirm on this Bootply](http://bootply.com/104843)

rhoggSugarcrm:
I'm working on a fix for this - it already works on Firefox and Chrome for my company's use cases. Testing for Safari and IE is in progress.

We need to have this fix into version 1.3.0, but we can fork that specific version ourselves if needed.

Message me if you have any questions. I haven't pushed my changes to my fork yet, but I will soon.

rhoggSugarcrm:
Ok, looks like my fix is working on FF, Chrome, Safari, IE 11, and IE 9. @acrobat are there any other browsers you would want me to test on?

My apologies that I can't post the code just yet - I need to make 100% sure I have legal permission - but expect it shortly (in a few days at the absolute worst).

acrobat:
Fixed in #1297

PR #1297, opened 2015-03-12
Enables right-click and edit-menu paste for the datepicker

rhoggSugarcrm:
Addresses #778 

Note: the test I added currently fails. The reason is that the expected UTC date is off:

>> Events - paste must update the date
>> Message: null
>> Actual: "2015-07-22 00:00:00.00"
>> Expected: "2015-08-22 00:00:00.00"
>> at datesEqual (file:///Users/rhogg/src/datePicker/tests/assets/utils.js:20)

Two existing tests fail on master (without any of my changes) for the same reason, so I'm not sure what to do. Either I could a) doctor the test so it passes and add a FIXME or b) leave it as is or c) fix the test because I'm doing something wrong, it's up to you.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -322,7 +322,8 @@
                     if ($.inArray(e.keyCode, [27, 37, 39, 38, 40, 32, 13, 9]) === -1)
                         this.update();
                 }, this),
-                keydown: $.proxy(this.keydown, this)
+                keydown: $.proxy(this.keydown, this),
+                paste: $.proxy(this.paste, this)
             };
 
             if (this.o.showOnFocus === true) {
@@ -476,6 +477,23 @@
 				delete this.element.data().date;
 			}
 			return this;
+		},
+
+		paste: function(evt){
+			var dateString;
+			if (evt.originalEvent.clipboardData && evt.originalEvent.clipboardData.types
+				&& $.inArray('text/plain', evt.originalEvent.clipboardData.types) !== -1) {
+				dateString = evt.originalEvent.clipboardData.getData('text/plain');
+			}
+			else if (window.clipboardData) {
+				dateString = window.clipboardData.getData('Text');
+			}
+			else {
+				return;
+			}
+			this.setDate(dateString);
+			this.update();
+			evt.preventDefault();
 		},
 
 		_utc_to_local: function(utc){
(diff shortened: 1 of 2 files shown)


rhoggSugarcrm:
Looking at Travis, it only shows my test failing and not the other two, so I'll post those failures here for reference:

>> Methods - setStartDate
>> Message: null
>> Actual: "2012-04-04 00:00:00.00"
>> Expected: "2012-04-05 00:00:00.00"
>> at datesEqual (file:///Users/rhogg/src/datePicker/tests/assets/utils.js:20)

>> Methods - setEndDate
>> Message: null
>> Actual: "2012-04-04 00:00:00.00"
>> Expected: "2012-04-05 00:00:00.00"
>> at datesEqual (file:///Users/rhogg/src/datePicker/tests/assets/utils.js:20)

acrobat:
@rhoggSugarcrm  the locally failing test could have to do with out-of-date dependencies (run npm install again) if that doesn't fix the problem just ignore those tests for now as they don't fail on travis or on my local checkout!

acrobat:
The changes look good to me! On what browsers did you test this change?

rhoggSugarcrm:
I can test on IE 8, expect an update later today.

acrobat:
Great thanks!

rhoggSugarcrm:
Looks like IE 8 works!

The same 3 tests still fail even after an npm update, still not sure what's up with them.

I noticed that the indentation is off, so I fixed that and pushed again

acrobat:
This failing test for this PR is caused by the wrong month parameter you pass to the `UTCDate` function. See this doc for more info https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC

> month
> An integer between 0 and 11 representing the month.

So if you update the test the following code, it should pass!

`datesEqual(this.dp.dates[0], UTCDate(2015, 6, 22));`

rhoggSugarcrm:
Ah, thank you! updated

acrobat:
Great thanks! Can you just squash the commits? Then the PR should be ready to merge!

rhoggSugarcrm:
done

acrobat:
Thanks @rhoggSugarcrm!

rhoggSugarcrm:
By the way, do you have an approximate ETA for the 1.4.1 release?

acrobat:
I have to check how much fixes we can get into that release but I would say around the end of april! (but that's a date that could change)

PR #1300, opened 2015-03-12
  Replace ESCAPE by DOWN key to re-show picker 2

fabdouglas:
Replace ESCAPE by DOWN key to re-show picker, rebase for #827


acrobat:

PR #1301, opened 2015-03-12
Replace ESCAPE by DOWN key to re-show picker (test)

fabdouglas:
Test update for  #1300.
Changes :
Escape when hidden --> does not show the picker anymore
Down when hidden --> show the picker


acrobat:

PR #1306, opened 2015-03-15
Allow selector in 'inputs' option for daterange

behigh:
Example
```
<div data-date-inputs=".datepicker" data-provide="datepicker">
   <input name=date_from class=datepicker> <input name="time_from">
   <input name=date_to class=datepicker> <input name="time_to">
<div>
```


acrobat:
Thanks for the PR @behigh, can you also update the docs for the inputs option?

PR #1307, opened 2015-03-16
Update dist with `grunt dist`.

korun:
Do not forget run `grunt dist` after merging, to update `dist` directory too!


acrobat:
Thanks korun for the PR but I do'nt think this should be done after each merge as the specific version branches and master branch are bleeding edge so the dist will be updated on release, if you use master or 1.4 branch you should use less files, and raw js. If you still want to use bleeding edge and want a dist file you should compile them yourself. 

I will leave this PR open to hear about others opinion but compiling the dist folder on each change and merge isn't something I want to do on development branches

korun:
Okay, I just tried to update timepicker PR to current master, and saw what `dist` has conflicts and not contain actual data. :)
So, now I should update PR without `dist` touching, right?

acrobat:
yes indeed! Dist conflicts can be ignored, just keep the versions of master in your case!

PR #1313, opened 2015-03-17
Highliting current date based on startDate arguments

sachitp:


js/bootstrap-datepicker.js
```diff
@@ -132,6 +132,9 @@
 		this.setEndDate(this._o.endDate);
 		this.setDaysOfWeekDisabled(this.o.daysOfWeekDisabled);
 		this.setDatesDisabled(this.o.datesDisabled);
+		if (options.startDate){
```
acrobat: I think you should replace the chech here with `o.startDate !== -Infinity` as the default value for the startDate is infinity

acrobat:
@sachitp Thanks for the PR but can you please fix the comment and revert the change in the dist file (this file will automatically be generated on release)

Also check test after you fixed the note! You can check/run the tests by running 
````
npm install
grunt test
```

vsn4ik:
Need to change `options.startDate` to `this.o.startDate`.

PR #1315, opened 2015-03-17
Feature : Days of week highlighted

Silhm:
I was looking for a feature that can do a bit more than daysOfWeekDisabled : Highlight some day(s) highlighting without disabling them. 
This is usefull for highlighting weekend for example.
The code is almost the same as the daysOfWeekDisabled feature, without the disabling part.

dayshighlighted


acrobat:
Cool feature @Silhm! Can you just add the docs for this option?

acrobat:
Thanks @Silhm!

Issue #1316, opened 2015-03-17
setDate() is not accepting a Date object.


trusktr:
For example, the following doesn't work, resulting in `getDate()` returning `null`.

```js
    $el.datepicker('setDate', new Date()); // should intuitively work
```

But the following *does* work.

```js
    $el.datepicker('setDate', (new Date()).toISOString());
```

I'm instantiating like this:

```js
    $el = $('[data-js=datepicker]').datepicker({
      format: 'mm-dd-yyyy',
      autoclose: true,
      endDate: endDate,
      todayBtn: "linked",
      todayHighlight: true
    });
```

and the markup looks something like this:

```html
  <div class="input-group date col-md-6" data-js="datepicker">
    <label for="date-input">Agent</label>
    <input id="date-input"
      name="date" class="form-control datepicker"
      type="text" placeholder="mm-dd-yyyy"/>
    <span class="input-group-addon"><i class="fui-calendar"></i></span>
  </div>
```

I've experienced this in v1.3.1 and v1.5.0. I haven't tried other versions yet.

trusktr:
In the following fiddle, it *does* work, but I'm not sure what version of datepicker it's using:

http://jsfiddle.net/DwT8M/413/

trusktr:
Here's a reproduction in jsfiddle, showing that today's date isn't selected (though the date *does* appear in the input unlike in my actual case):

http://jsfiddle.net/DwT8M/415/

trusktr:
Here's a reproduction in jsfiddle, showing that today's date isn't selected (though the date does appear in the input unlike in my actual case):

http://jsfiddle.net/DwT8M/416/

trusktr:
Okay, so calling `update()` does update the selection, so that's fine and my (client's proprietary) code is working now. Upgrading from 1.3.1 to 1.5.0 seems to have fixed a problem where the `changeDate` wasn't always firing after calling `setDate`. Anyway, it's all working now.

acrobat:
Good that the problem is solved @trusktr, but keep in mind that 1.5.0 is still in development! 1.4.0 is the latest stable release

Issue #1308, opened 2015-03-16
Bower Pulling Old Version


prionkor:
I just installed the plugin via bower. It is pulling version `1.4.0`. Any idea why?

acrobat:
1.4.0 is the latest released version, which version did you expect to be pulled by bower?

prionkor:
Some of files says 1.5.0 or it isn't stable yet?

https://github.com/eternicode/bootstrap-datepicker/blob/master/dist/js/bootstrap-datepicker.min.js
https://github.com/eternicode/bootstrap-datepicker/blob/master/bower.json

acrobat:
no 1.5.0 is the current development version, 1.4.0 is the latest released version

This is what the `bower info bootstrap-datepicker`command gives as output

```
Available versions:
  - 1.4.0
  - 1.3.1
  - 1.3.0
  - 1.3.0-rc.6
  - 1.3.0-rc.5
  - 1.3.0-rc.4
  - 1.3.0-rc.3
  - 1.3.0-rc.2
  - 1.3.0-rc.1
  - 1.2.0
  - 1.2.0-rc.1
  - 1.1.3
  - 1.1.2
  - 1.1.1
  - 1.1.0
  - 1.0.2
  - 1.0.2-rc.2
  - 1.0.2-rc.1.1
  - 1.0.2-rc.1
  - 1.0.1
  - 1.0.0
```

So you should require 1.4.0 in your bower config

prionkor:
thank you. I thought something was wrong..

PR #1318, opened 2015-03-17
change version of master

hebbet:
after comments in #1316 and #1308 i think some people are confused about the current stable release, i suggest to change the version number in master from 1.5.0 to 1.5.0-dev


acrobat:
It's indeed a good idea to add the dev suffix to avoid confusion! Thanks @hebbet

korun:
Why not `rc-N` (release candidate) as before?

acrobat:
We can use the RC suffix if we are almost ready with 1.5.0, tag an RC version and have it out so people can test this new version in their apps. For now -dev is ok for the version in master

PR #1327, opened 2015-03-20
Add immediateUpdates option (updates input when a year or month is selected)

### LINKED ISSUE #1323: Add an option to automatically update the date field when a year or month is selected
I am working on an application which uses bootstrap-datepicker for a form field asking for a user's birthday. In some cases, a user could try to edit a date they have already entered by updating the year or month in the datepicker. However, this does not save the new date to the input field until the user chooses the date again, which means that it is possible for the user to not realize that their change did not occur and cause them to submit the wrong date.

I have created [an example](http://codepen.io/nicolasmccurdy/pen/RNEboj) which uses the datepicker's events to force input field updates when the year or month change. I think it would be useful if there was an option to enable this behavior built into the datepicker.

nicolasmccurdy:
Fixes #1323.

> immediateUpdates
> ---------------------
> 
> Boolean. Default: false
> 
> If true, selecting a year or month in the datepicker will update the input value immediately. Otherwise, only selecting a day of the month will update the input value immediately.

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -297,3 +297,10 @@ enableOnReadonly
 Boolean. Default: true
 
 If false the datepicker will not show on a readonly datepicker field.
+
+immediateUpdates
+---------------------
+
+Boolean. Default: false
+
+If true, selecting a year or month in the datepicker will update the input value immediately. Otherwise, only selecting a day of the month will update the input value immediately.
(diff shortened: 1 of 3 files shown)


acrobat:
Thanks @nicolasmccurdy!

Azaret:
I noticed that when selecting a month or a year, the day is not kept and set to 1.

Issue #1233, opened 2015-01-13
datepicker `change` event fired twice.


kntmrkm:
When 'change' event set to datepicker field.

'change' event fire twice.

In repo [vizjerai/bootstrap-datepicker](https://github.com/vizjerai/bootstrap-datepicker/).

This issue is fixed like this.

[Fix multiple calls to change #4](https://github.com/vizjerai/bootstrap-datepicker/issues/4)
[commit](https://github.com/vizjerai/bootstrap-datepicker/commit/db20cae1d6ada09c9693fb780567388508b3ff57)

How is going at this repo?

This fixing is correct?

acrobat:
Yes I saw that patch before, I'm going to look at that issue soon and see if the patch is correct!

kntmrkm:
OK. Thanks.

mgodwin:
:+1:

acrobat:
Fixed in #1337

akhayoon:
Does this mean the issue was fixed in 1.5.0? 

This is because I'm using 1.6.0 and I'm still getting the event triggered 3 times on the 'change dp.change' event. Please advise if I'm doing something wrong. Thank you

seanjamu:
this problem persist now in range datepicker, when select date in a input it cause two events

vqminh:
Still see this on 1.6.1

acrobat:
Can you guys provide a jsfiddle to show if the problem still persists? It should be fixed with the pr, but maybe some new code broke it against

yassirtaqatqa:
Just change it to...
`.on('changeDate',function() {
                    do_staff();
                });`

mistrynilesh:
I am using 1.6.4.  Still facing same issue in version on 1.6.4.

roysupriya:
In my case, I am using daterangepicker with angular. My purpose is to watch any change in the model that stores the date range value and save it for an AJAX call later. I face the same issue as it hits the event twice whenever the date is change even if it is only 'Today': **_Once it is object with startDate and endDate properties, and the other time it is a string._**

**It can be made use of as an advantage.**    
```
$scope.$watch(
    'rangeOfDate',
    function (newValue) {
        // Due to a known bug of open source library daterangepicker, the event is hit twice 
        //upon change. Once it is an object, and once it is a string. So, use appropriately.
        var selectedDateRange = new Object();
        if (typeof (newValue) == 'object') {
[...]

PR #1337, opened 2015-03-25
Fixes the issue with a date change firing mulitple change events

darylldoyle:
This PR fixes the issue with a date change firing multiple change events as reported in #1233 

Fix is pulled from [this issue](https://github.com/vizjerai/bootstrap-datepicker/issues/4) and [this commit](https://github.com/vizjerai/bootstrap-datepicker/commit/db20cae1d6ada09c9693fb780567388508b3ff57)

I've tested this locally and all seems to be working OK with it.

### DIFF
--- a/tests/suites/events.js
+++ b/tests/suites/events.js
@@ -301,7 +301,7 @@ test('setDate: triggers change and changeDate events', function(){
 
     this.dp.setDate(new Date(2011, 2, 5));
 
-    equal(triggered_change, 0);
+    equal(triggered_change, 1);
     equal(triggered_changeDate, 1);
 });
 
--- a/tests/suites/events.js
+++ b/tests/suites/events.js
@@ -301,7 +301,7 @@ test('setDate: triggers change and changeDate events', function(){
 
     this.dp.setDate(new Date(2011, 2, 5));
 
-    equal(triggered_change, 2);
+    equal(triggered_change, 0);
     equal(triggered_changeDate, 1);
 });
 
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -773,6 +773,7 @@
 				this._trigger('clearDate');
 
 			this.fill();
+			this.element.change();
 			return this;
 		},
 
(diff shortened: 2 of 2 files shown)


hebbet:
thanks. could you also fix the test?

acrobat:
When the PR is ready we should merge this in the 1.4 branch, so we can push this fix the next bugfix release

darylldoyle:
No worries, will update it tonight when I'm home!

tests/suites/events.js
```diff
@@ -301,7 +301,7 @@ test('setDate: triggers change and changeDate events', function(){
 
     this.dp.setDate(new Date(2011, 2, 5));
 
-    equal(triggered_change, 2);
+    equal(triggered_change, 0);
```
acrobat: Should the change event not be called once? Or do we just want to leave out the change event and only have a changeDate event, but that would mean a BC break

darylldoyle: The reason for taking it out completely is that the change event will be fired on the base element anyway won't it? Therefore if someone is targeting the base element with an `on.('change', ...)` handler, it will always be fired more than once. 

On the other hand, if someone want to specifically listen for a date change, they can listen for the `changeDate` event.

Would it be worth moving this PR to target a 2.0 branch if using semver or otherwise 1.4 to account for the BC break?

acrobat: Looking at the result of the test, the change event isn't fired anymore on the input field, if we just fire change once it wouldn't be a BC break. The issue is that it fires the change event more then once, if it would just fire once this bug is fixed without the BC break

darylldoyle:
I think this may solve it. I did a bit of profiling and it seems that the `update` method is called pretty much anytime something is done through the public API. Therefore I've removed the change event from the `clearDates` method and added it to the bottom of the `update` method.

You may have more insight on whether this will or wont work though!

vsn4ik:
:+1: for merge.

darylldoyle:
Any news on this one?

acrobat:
Looks good @darylldoyle! Thanks!

PR #1340, opened 2015-03-27
Better Polish language date shortcuts translation and default date format

crabonature:
I would like to propose better from Polish grammatical point of view language translation for weekdays, months and their shortcuts.

Added default date format.


acrobat:
Thanks @crabonature!

Issue #1342, opened 2015-03-28
 in-line datepicker disappears on blur.


acrobat:
This issue was reported by email

---

The latest changes to the datepicker cause the in-line calendar to disappear on blur.

At line 381
  this.hide();
has been replaced by:
  $(this.picker).hide();

The internal hide() returns if the datepicker is inline.

Regards
Ivano Vingiani

---

This bug was introduced with PR #1322

- [ ] find solution to fix this issue but keep #1145 also fixed
- [ ] add test to prevent future regression

3rdPancake:
how about this on line 381 ?
```javascript
this.isInline ? true : $(this.picker).hide();
```
not sure about returning true, but this can prevent from hiding inline datepicker on blur.

acrobat:
Yeah that looks like a possible fix, I will test it later and if it works add a test for it to prevent future regression

trinzia:
This is probably same as my issue #1339.

acrobat:
Looks like this issue is fixed in #1399

Issue #1145, opened 2014-11-03
Prototype Compability


EliasKotlyar:
In this Function
```
this._secondaryEvents = [
				[this.picker, {
					click: $.proxy(this.click, this)
				}],
				[$(window), {
					resize: $.proxy(this.place, this)
				}],
				[$(document), {
					'mousedown touchstart': $.proxy(function(e){
						// Clicked outside the datepicker, hide it
						if (!(
							this.element.is(e.target) ||
							this.element.find(e.target).length ||
							this.picker.is(e.target) ||
							this.picker.find(e.target).length
						)){
						this.hide();
						}
					}, this)
				}]
```
There is an Error when using with Prototype. Correct it by changing it to
```
this._secondaryEvents = [
				[this.picker, {
					click: $.proxy(this.click, this)
				}],
				[$(window), {
					resize: $.proxy(this.place, this)
				}],
				[$(document), {
					'mousedown touchstart': $.proxy(function(e){
						// Clicked outside the datepicker, hide it
						if (!(
							this.element.is(e.target) ||
							this.element.find(e.target).length ||
							this.picker.is(e.target) ||
							this.picker.find(e.target).length
						)){
						$(this.picker).hide();
						}
					}, this)
				}]
```

rainmaker2k:
This fix actually breaks when using datepicker inline with jQuery. The datepicker will always be hidden, because it will always call jQuery's hide instead of Datepicker's own hide method.

davilima6:
+1 @rainmaker2k This is closing datepicker if I click wherever else on the page. Please revert or provide an alternative fix with tests, please.

razvanphp:
+1

other duplicates: #1339 and #1366, you might want to close them as cleanup.

MrKuja:
The bug still occurs on latest master branch.
Quite annoying on such pleasant library.

pashkasan:
still not fixed!

PR #1348, opened 2015-03-30
Trigger "hide" event on click outside

WojtekKruszewski:
`hide` event is being triggered correctly when popup is dismissed with escape key, by tabbing to next field or clicking date or clear button. This change makes it trigger also popup is dismissed by clicking outside.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -394,7 +394,7 @@
 							this.picker.is(e.target) ||
 							this.picker.find(e.target).length
 						)){
-							$(this.picker).hide();
+							this.hide();
 						}
 					}, this)
 				}]


acrobat:
This line was recently changed by PR #1322 which backported a fix from #1145. So maybe we can get some feedback from @EliasKotlyar why he suggested the fix and how we can come up with a workaround

/cc #1342

p5yk0:
When you use the calendar as a multi monthpicker , the `hide` event isn't triggered when clicking outside.
I've fixed it the same manner as WojtekKruszewski  did it.

davilima6:
+1 for this pull

acrobat:
@EliasKotlyar Can you give some feedback on this issue and on how to fix this issue? Without having that your issue/fix is reverted (#1145)

/cc #1342

acrobat:
ping @EliasKotlyar

EliasKotlyar:
Hello,

I was facing this issue(https://github.com/eternicode/bootstrap-datepicker/issues/1145) when using the datepicker in Magento. Magento has prototype as its primary JS-framework. The prototype-library was interpreting the "this.hide()" command with its own implementation of "hide()" instead of using the function of the datepicker. Therefore i fixed it up by using a jQuery function. This fix is not needed except you are using Prototype and jQuery together.

Best Regards
Elias

acrobat:
@EliasKotlyar This looks more like an issue of prototype instead of the datepicker. I'm going to merge this fix and have a look later at usage of the datepicker with prototype and jquery. In the meantime can you provide a jsfiddle with jquery and prototype to reproduce the issue later?

acrobat:
Thanks @WojtekKruszewski!

PR #1369, opened 2015-04-14
Fix duplicated code introduced in affed43e461106afcf4dcaaf500fe0763dbf5f88

nmeylan:
Hello,

It seems you missed something here : https://github.com/eternicode/bootstrap-datepicker/blob/master/js/bootstrap-datepicker.js#L214

This PR should fix it, I ran the tests suite, nothing failed, but I'm not sure this part is covered.

Have a nice day.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -212,7 +212,6 @@
 			}
 
 			o.startView = Math.min(o.startView, o.maxViewMode);
-			o.startView = Math.max(o.startView, o.minViewMode);
 
 			// true, false, or Number > 0
 			if (o.multidate !== true){


acrobat:
Thanks for the PR, I'm gonna figure out what has caused this duplicate line and what it should be! Will reply later on to this PR when I've found the cause!

nmeylan:
Hi,

Sorry I was wrong, I deleted the wrong line, I kept the commit author intention but it was wrong.

Azaret:
There is nothing wrong there, both lines are used to keep `startView` in range between `minViewMode ` and `maxViewMode`.
This PR should not be merged and closed.

hebbet:
Ping @acrobat  ;-)

hebbet:
1 Packet lost. try again ;-)

ping @acrobat  :-)

acrobat:
oh sorry :D I guess @Azaret is right, as the 2 lines do some different logic

PR #1385, opened 2015-04-25
Make data attributes case-insensitive and add 'today' as option for startDate and endDate.  Add DPGlobal to allow date formatting in events.  Also some spaces/tabs cleanup.

mreiden:


mreiden:
Without this, data attributes end up adding unused attributes enddate and startdate instead of modifying endDate and startDate.

PR #1367, opened 2015-04-13
Updates for zh-cn/zh-tw.  Korean language support.

hemp:
Our translation team prefers the usage of [month number]月 for month selection as this is more commonly seen (one big example is the Windows date-time picker).  This also allows better spacing for November and December which are 3 characters long and they run together.

Solution: Switch `monthsShort` for zh-cn/zh-tw to use `monthsShort` from moment.js.  Add Korean language support by porting from moment as well.  In addition a `titleFormat` key has been added for the `.datepicker-switch` header of the date picker.  This text used to be created with string concatenation and did not work well for Chinese and Korean.  Chinese and Korean provide a `titleFormat` key override and this key defaults to `"MM yyyy"` for English and other languages.

PR #1396, opened 2015-05-13
locale: 'kr' is not language code

hiphapis:
'kr' is not language code. it's country code.
Korean language code is 'ko'
so i did rename to 'ko'

### DIFF
--- a/js/locales/bootstrap-datepicker.ko.js
+++ b/js/locales/bootstrap-datepicker.ko.js
@@ -10,6 +10,4 @@
 		months: ["1월", "2월", "3월", "4월", "5월", "6월", "7월", "8월", "9월", "10월", "11월", "12월"],
 		monthsShort: ["1월", "2월", "3월", "4월", "5월", "6월", "7월", "8월", "9월", "10월", "11월", "12월"]
 	};
-  // for compatibility
-  $.fn.datepicker.dates['kr'] = $.fn.datepicker.dates['ko'];
 }(jQuery));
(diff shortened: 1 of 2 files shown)


acrobat:
To not break compatibility we should also keep the kr file, but add a deprecation notice at the top of the file (so we can remove it in 2.0)

Thanks!

js/locales/bootstrap-datepicker.ko.js
```diff
@@ -3,11 +3,13 @@
...
-	$.fn.datepicker.dates['kr'] = {
+	$.fn.datepicker.dates['ko'] = {
 		days: ["일요일", "월요일", "화요일", "수요일", "목요일", "금요일", "토요일"],
 		daysShort: ["일", "월", "화", "수", "목", "금", "토"],
 		daysMin: ["일", "월", "화", "수", "목", "금", "토"],
 		months: ["1월", "2월", "3월", "4월", "5월", "6월", "7월", "8월", "9월", "10월", "11월", "12월"],
 		monthsShort: ["1월", "2월", "3월", "4월", "5월", "6월", "7월", "8월", "9월", "10월", "11월", "12월"]
 	};
+  // for compatibility
+  $.fn.datepicker.dates['kr'] = $.fn.datepicker.dates['ko'];
```
acrobat: If the original ko file is restored you can remove this line

hiphapis:
Great news!
yeah. i restored that line
Thanks ;)

acrobat:
Great, please just restore the original ko file to (for people currently using the ko language code. But also keep the new kr language file just so that we have a alternative when we deleted in 2.0 (and add some deprecation comment in the old ko file)

hemp:
#1367 spruced up support for ko/zh.

hiphapis:
@acrobat Ah. you mean is restore legacy file. right?

acrobat:
Thanks for the pr @hiphapis, but #1367  got merged. So closing this!

PR #1399, opened 2015-05-14
Do not hide inline picker

alsofronie:
When in a custom setup (I have two inline pickers in a bootstrap collapse div), the  click event from the collapse trigger will set `style="display:none"` on the two inline pickers. Testing for the `.datepicker-inline` class on the picker will inhibit this (unwanted) behaviour.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -420,7 +420,8 @@
 							this.element.is(e.target) ||
 							this.element.find(e.target).length ||
 							this.picker.is(e.target) ||
-							this.picker.find(e.target).length
+							this.picker.find(e.target).length ||
+							this.picker.hasClass('datepicker-inline')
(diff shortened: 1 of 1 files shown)


rexromae:
To fix this issue, the lines of code 419-426 of the file bootstrap-datepicker.js should be replaced with the following code.

```
						if (!(
							this.element.is(e.target) ||
							this.element.find(e.target).length ||
							this.picker.is(e.target) ||
							this.picker.find(e.target).length ||
                                                        this.isInline
						)){
							$(this.picker).hide();
						}
```

alsofronie:
Check my commit, it's exactly that :)

rexromae:
You're right, I've seen it only after commenting. Has your commit been merged with the master repository?

acrobat:
Thanks @alsofronie!

PR #1400, opened 2015-05-15
'container' update

MikeSha:
Datepicker dropdown will be added to the closest element pointed in 'container' option. So, you can save some bytes if you have several datepickers in one page :)

This was useful for me :)


hebbet:
@MikeSha Can you please take a look at travis and fix the build issue?

hebbet:
closing old tickets, with no newer reply.

PR #1417, opened 2015-05-28
point the bower package at the dist/*.js rather than js/*.js

ETNOL:
The bower package is pointing at the pre-build bootstrap-datepicker.js.  This changes it to point at the dist version.


acrobat:
Thanks @ETNOL!

PR #1419, opened 2015-06-02
Add ability to retrieve z-index from stylesheet

CheapSk8:
PR 1242 updated against current master.
----------
Updated z-index functionality in "place" function to check for explicit z-index set through stylesheet. Current func. retrieves z-index based on DOM stack. In some instances you might want to specify the z-index through a CSS class instead. Updated func. will first reset picker z-index and check for z-index set via stylesheet. If not-specified, DOM stack is used. *Corrected conditional for line 596 zIndex var.


acrobat:
@CheapSk8 thanks for the effort! But can you take a look at the travis build, there are some jscs issues and can you maybe add a test to check this new behavior of the z-index checker. That way we can avoid future regressions!

CheapSk8:
You will have to forgive me for my ignorance to this, but I am still fairly new to working with GitHub as a repo I submit to. How would I add a test for this?

acrobat:
No problem @CheapSk8! You can take a look at these test https://github.com/eternicode/bootstrap-datepicker/blob/master/tests/suites/component.js, you can use jquery's functions to check the css properties

rosskevin:
I have a situation where a sibling element (not parent) is absolutely positioned.  In this rare case, hardcoding the z-index with `!important` works i.e.

```css
.datepicker-dropdown {
  z-index: 1036 !important;
}
```

Since `!important` works, it could be argued that this pull may not be needed.  On the other hand, many people don't know about `!important`, so this pull request is helpful.   Either way, I wanted to note my current workaround for the case of overlapped absolutely positioned components that are not in the same stack.

PR #1422, opened 2015-06-05
Moved jquery to peerDependencies

Semigradsky:


acrobat:
Hi @Semigradsky,

Can you explain a bit more what the difference is between `dependencies` and `peerDependencies`?

PR #1428, opened 2015-06-14
fix changeMonth, changeYear not triggered when selecting day from prev/next month

r0ro:


acrobat:
@r0ro can you rebase the branch against the current master?

r0ro:
I'll try to do that asap

r0ro:
Done.
Note that I'm not sure it sufficient for every cases, for instance if the user clicks on 'today' should the 'changeYear', 'changeMonth' trigger ?

acrobat:
In my opinion it should not trigger when clicking the today button (could be a seperate event). Will do a final review soon and have it merged for 1.5.1. This now fixes cases that were broken, we can fix other cases when they are reported!

acrobat:
Thanks @r0ro!

PR #1430, opened 2015-06-16
Fix #1406

### LINKED ISSUE #1406: Datepicker orientation mix up - top = bottom and bottom = top
I tried to set the orientation of my datepick so that it always goes above the field like so - 

$('.dateInput').datepicker({
            language: "en-GB",
            orientation: "top auto"
        });

This did not work, the datepicker would always show up on the bottom until I changed the code to this -

$('.dateInput').datepicker({
            language: "en-GB",
            orientation: "bottom auto"
        });

Azaret:
This should fix #1406 if we aim to make more sense with top and bottom.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -729,9 +729,9 @@
 			}
 			this.picker.addClass('datepicker-orient-' + yorient);
 			if (yorient === 'top')
-				top += height;
-			else
 				top -= calendarHeight + parseInt(this.picker.css('padding-top'));
+			else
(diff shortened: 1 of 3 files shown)


hebbet:
LGTM, merged, thanks

Azaret:
By the way, I didn't run any grunt or less stuff, don't know if you want to run it to regenerate the dist folder.

acrobat:
The dist folder will be re-generated when we release a new version!

Azaret:
Good !

Issue #1429, opened 2015-06-16
border-radius in less


BARTlab:
pls, replace 
```border-radius: 4px;```
to
```border-radius: @border-radius-base;```

hebbet:
feel free to make a PR.

Azaret:
I felt free, that said, I think that some day the LESS files would need some cleaning, naming convention are different from one file to another.

acrobat:
Fixed in #1435

PR #1435, opened 2015-06-17
Adding border-radius variable for LESS

Azaret:
PR for #1429

### DIFF
--- a/less/datepicker3.less
+++ b/less/datepicker3.less
@@ -1,5 +1,5 @@
 .datepicker {
-	.border-radius(@border-radius-base);
+	border-radius: @border-radius-base;
 	&-inline {
 		width: 220px;
 	}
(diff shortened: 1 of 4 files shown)


less/datepicker3.less
```diff
@@ -1,5 +1,5 @@
 .datepicker {
-	border-radius: 4px;
+	.border-radius(@border-radius-base);
```
acrobat: Hmm i'm not exactly sure why you added the 'dot' here? Or why it is in the BS2 file, should it be a "normal" property without the dot?

Azaret: Well the mixin is used on the other file, I guessed it made sense to use it here as well.
Now I see that there is no such mixin on the build, gonna fix this.

acrobat: Oh yes i forgot about the mixins! Thanks for the correction!

acrobat:
Thanks @Azaret!

Issue #1226, opened 2015-01-09
Adding CSS classes to years (viewMode 2) doesn't work


sth519:
Adding CSS classes to certain years, for example using the beforeShowDay or onRender callbacks, doesn't work with year selections (minViewMode 2). The functions get called but return values have no effect.

acrobat:
@sth519 please add a jsfiddle example showing the case, makes it easier to debug ;) You can use [this jsfiddle](http://jsfiddle.net/g02L8z8y/) as a base to create your case!

sth519:
Simplified example: http://jsfiddle.net/ro9ufa3k/1/

If you comment out the view mode settings, everything works as documented. However, I have not found a way to assign classes in the year selection mode.

sth519:
Any ideas on how to fix this? I'm not familiar with the codebase of this project, so any pointers in the right direction are highly appreciated.

Thanks!

sth519:
Any ideas?

acrobat:
The option you use in the example is the before show day callback and it only applies to the day viewmode. There is another option 'beforeShowMonth' but there is not one yet for the year view

So if I understand you correct, what you try to achieve needs an new option 'beforeShowYear'?

sth519:
Yes. The reason is that I need to highlight certain years in the dropdown list.

acrobat:
Ok, I will label this issue as an feature and have it go into the 1.5.0 release!

sth519:
Thanks!

acrobat:
Fixed in #1436

PR #1436, opened 2015-06-18
Implement `beforeShowYear` as requested in #1226

Azaret:
I reproduced `beforeShowDay` behavior.


hebbet:
LGTM.

acrobat:
@Azaret can you just a test for this new option? You can use the beforeShowDay option test to copy from!

Azaret:
@acrobat If everything is ok with this PR can we merge it ? I'm afraid it will conflict a bit with #1437

acrobat:
Great, thanks @Azaret!

Issue #1407, opened 2015-05-22
immediateUpdates option does not seem to work


bensinclair:
The `immediateUpdates` option does not seem to work. Have I misunderstood something?

Here's a demo: https://jsfiddle.net/wbLunk0j/

Azaret:
Looks fixed in 1.5

Azaret:
Actually this is a 1.5 feature, the JS you linked is 1.4.

bensinclair:
Thanks mate! I thought I was going crazy!

+1 to showing stable and not dev docs :) Or at least put a comment in about what version is required for a specific feature.

PR #1440, opened 2015-06-18
Link to last stable release docs instead of dev docs

Azaret:
@acrobat I think that the readme should lead people to the last stable docs instead to the dev docs. 
It might help avoid misunderstanding like in #1407.


hebbet:
merged.

PR #1359, opened 2015-04-06
Fix daysOfWeekHighlighted with startDate (endDate).

vsn4ik:

PR #1447, opened 2015-06-24
Removed erroneous day highlighting after endDate

alexserdyuk:
When you set startDate or endDate value, disabled days become
highlighted. Bug was added in commit
77ea1ea1d3cfd4a44db88b63fe966f9a0bdf3919


vsn4ik:
Already exist PR (#1359).

vsn4ik:
Close by merge #1359.

acrobat:
This PR has an extra test maybe we should port this test?

PR #1448, opened 2015-06-24
added example to end-date

stenmuchow:
To avoid future confusion of forgetting to place a 'd' after the 0.


acrobat:
Thanks @stenmuchow!

PR #1451, opened 2015-06-25
Add commonjs support

zaynetro:


vsn4ik:
:+1: for merge.

acrobat:
Thanks @zaynetro!

vshulev:
Hi,
I'm trying to use bootstrap-datepicker with browserify. I get a `$(...).datepicker is not a function` error. At the top of my file I have:
```
var $ = require('jquery');
require('bootstrap-datepicker');
```
Is this the proper way to use it?

vsn4ik:
Use `require('bootstrap-datepicker', $);`

vshulev:
@vsn4ik, thanks for the reply! This also doesn't seem to work (tried it and got an error). The issue seems to be that nothing is added to `module.exports` and therefore `require` returns just an empty object. I've created a PR for this - #1703, after the changes the syntax you are proposing - `require('bootstrap-datepicker')($);` works fine.

zaynetro:
Could you provide an example when this breaks? I use bootstrap date picker in production and it does work with browserify. Perhaps there might be different between versions in how jquery/npm/browserify manage modules.

vshulev:
I'm using the bootstrap datepicker inside an Angular directive (not sure whether this makes any difference). These are the top 3 lines in my directive file: 

```
var moment = require('moment');
var $ = require('jquery');
require('bootstrap-datepicker');
```

I also tried `require('bootstrap-datepicker')($)` - doesn't work either. The datepicker plugin doesn't get attached to jQuery, and so I get an element is undefined error when I try to initialize an element. This is the code for initializing the datepicker:

```
$(elem).datepicker({
        format: 'dd/mm/yyyy',
        startDate: '-3d',
        autoclose: true,
        startDate: '-100y',
        endDate: '0d',
        startView: scope.startView,
        forceParse: false
});
```
It used to work before I ported the code to browserify. I'm using the following versions:
- browserify 12.0.1
- bootstrap-datepicker 1.5.0
- jQuery 2.1.4

zaynetro:
@vshulev, take a look at this gist https://gist.github.com/zaynetro/b575f0dc428bbbf9c810 . Example there works with out any troubles.

vshulev:
@zaynetro, you're absolutely right! Tried your example, then tried the same thing I was trying yesterday and now it works without a problem... Not quite sure what the issue was, but closing the PR. Thanks for the gist!

PR #1453, opened 2015-06-25
Update json accordingly the latest definition

GabrielGil:
Bootstrap has done it. We as well. Update the bower.json file to only include one line per file type.
Only the main CSS file is listed then.

Ref: https://github.com/bower/bower.json-spec/pull/43


acrobat:
Thanks @GabrielGil!

Issue #1457, opened 2015-06-25
Feeding 2-digit years into yyyy formatted datepicker


kremerd:
There seems to be a problem when entering two digit years in a datepicker input box that is configured to use the four digit format (like the default mm/dd/yyyy).

When I select a datepicker input box (like the one on http://eternicode.github.io/bootstrap-datepicker), enter 01/01/01 by keyboard and push Enter, the following happens depending on the browser used:
* With IE 11 the input box value is changed to 25/06/1 and in the datepicker widget 25/06/2015 (today) is still selected.
* With Firefox 38 the input box value is changed to 25/06/01 and in the datepicker widget 25/06/2015 (today) is still selected.
* With Chrome 43 the input box value is changed to 01/01/1 and in the datepicker widget 25/06/2015 (today) is still selected. However, the displayed calendar page is changed to "January 1".

Neither of these results seem to be appropriate, although Chrome is getting close. I would expect that the value in the input box changes to either 01/01/1, 01/01/2001, or (at least) not at all. In any case the date selected in the widget should match the string in the input box and the appropriate calendar page should be displayed.

PR #1461, opened 2015-06-26

asessa:
If the specified format is yyyy but user enters an year using other digits, this is converted to 2XXX to respect the yyyy format.

f.e. 1 > 2001, 15 > 2015

Issues: #1457 #1446

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1727,6 +1727,7 @@
 				setters_order = ['yyyy', 'yy', 'M', 'MM', 'm', 'mm', 'd', 'dd'],
 				setters_map = {
 					yyyy: function(d,v){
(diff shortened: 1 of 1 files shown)


asessa:
Tests should be updated too to respect the specified date format

Invalid formats are force-parsed into a valid date on tab:

equal(this.input.val(), '2056-September-30');

Azaret:
Many people would expect 90's years to be converted to the 20th century, maybe even the 80's...
Where to draw a line ?

To be honest, I am totally against this.
Everyone agrees that `setYear` allowing to change year 0-99 to 1900-1999 was a mistake, MSDN and MDN discourage it usage, Chrome already removed this behavior.

This could be an option maybe, but definitely not a default behavior.

PR #1465, opened 2015-06-30
feat(require): make requirable by adding index.js

jiaweihli:
This allows for requiring bootstrap-datepicker from other files using the syntax `require('bootstrap-datepicker')`, which will pull in the unminified source.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,2 @@
+$ = jQuery = require('jquery');
+module.exports = require('./dist/js/bootstrap-datepicker.js');


jiaweihli:
Avoided adding dependency on `bootstrap` since npm versions start at 3.1.1.
However, jquery npm versions start at 1.5.1, which I think is 'old enough'.

I can add a comment to `index.js` re: bootstrap omission if you prefer.

acrobat:
Isn't this already possible by [this lines](https://github.com/eternicode/bootstrap-datepicker/blob/master/js/bootstrap-datepicker.js#L26)?

jiaweihli:
This was fixed by https://github.com/eternicode/bootstrap-datepicker/commit/bdf7658f2195b28373e7de177a0e2acd4fbc357e, which was released in v1.5.0 in October.

PR #1464, opened 2015-06-30
Trigger 'hide' event when clicking outside datepicker

alexserdyuk:

PR #1469, opened 2015-07-01
fix hiding

mreinsch:
previously clicking on an element outside of the datepicker didn't go through the usual hide method, thus forceParse was ignored, leaving invalid dates in the input field

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -425,8 +425,7 @@
 							this.picker.find(e.target).length ||
 							this.picker.hasClass('datepicker-inline')
 						)){
-							$(this.picker).hide();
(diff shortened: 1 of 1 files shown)


hebbet:
possible duplicate of #1464

mreinsch:
@hebbet, #1464 didn't fix that issue though. As you can see, my changes are on top of master, which includes #1464. 

I've confirmed with master, that the issue is still there:

* change the date field to have invalid date
* click somewhere outside the date picker
* the field will keep the invalid value

On the other hand if you use 'tab' key to jump to next field, the invalid date will get replaced with a valid date.

mreinsch:
thanks! :)

PR #1471, opened 2015-07-02
added setDatesDisabled method to documentation

mikedebock:
I was looking for a method to update the datesDisabled value.
Until I discovered that it had already implemented but not documented yet.


acrobat:
Great thanks @mikedebock!

PR #1478, opened 2015-07-10
fix picker layout and orientation when rtl is true and input box is wide

Ahed91:
Hi ,, 
There is a problem with datepicker and rtl option which result in white space next to the picker and orientation not working properly
check this jsfiddle 
    http://jsfiddle.net/Ahed91/g18gry8a/3/


rizwanpervaiz:
+1 for this please

PR #1481, opened 2015-07-13

helldone1986:
added short year limit.
if setting year > 20 then date 19xx else 20xx


hebbet:
not 100% sure what the benifit of that change is.

acrobat:
#1556 does also fix this "issue" but I think that PR is a better solution

PR #1483, opened 2015-07-13
Use date arithmetic to allow dates with times other than 00:00:00

### LINKED ISSUE #1221: setDate does not work unless time is 00:00:00
When using setDate on an inline datepicker, the selected date is not updated unless the date's time is 00:00:00.

Does not work:
$("#datepick").datepicker('setDate',new Date("2015-02-13 09:30:00"))

Works:
$("#datepick").datepicker('setDate',new Date("2015-02-13 00:00:00"))

Here is a jsfiddle demonstrating the issue:
http://jsfiddle.net/Loe6kfwt/1/

colavitam:
Instead of directly comparing date values using valueOf, dates are now compared with arithmetic to check if they fall under the same date. This fixes resolves #1221, allowing for times other than 00:00:00 when setting the date.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -60,7 +60,8 @@
 				// $.inArray doesn't work with Dates
 				var val = d && d.valueOf();
 				for (var i=0, l=this.length; i < l; i++)
-					if (this[i].valueOf() === val)
+          // Use date arithmetic to allow dates with different times to match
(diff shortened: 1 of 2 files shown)


acrobat:
@Azaret What do you think? Fix looks good enough to me as a temp fix until we support time picking!

Azaret:
Looks good to me!

acrobat:
Thanks @colavitam

PR #1488, opened 2015-07-15
Fix bug on Range

WedgeSama:
If you want to init a datepicker (range or not) after remove a range one on the same input(s), an error was throw.

The case is if you want to dynamically change a range picker to a solo picker.

To test with the [online demo](http://eternicode.github.io/bootstrap-datepicker), choice range then run that in the console:
```js
$('#datepicker').datepicker('remove');
$('#datepicker input[name=start]').datepicker({});
```

I try many cases with remove (included call remove on each range children).

[EDIT] Forgot to say, the error is throw only if a value as been add to the input before removing range.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1494,8 +1494,13 @@
 				return;
 			this.updating = true;
 
-			var dp = $(e.target).data('datepicker'),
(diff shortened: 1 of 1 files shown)


acrobat:
Thanks @WedgeSama!

PR #1493, opened 2015-07-16
Add the zIndexOffset option to configure the dialog "height"

laserlemon:
Previously, the height was computed to be 10 higher than the max height of all of the input's parents. In some cases where the datepicker needs to move/scroll around with the page, it may collide with other fixed elements. In those cases, a configurable z-index is very helpful in order to keep the datepicker visible.

Please let me know if I should also be regenerating the `dist` assets as well. I couldn't infer from the contribution guidelines. Thank you!

CSS

## TODO

* [x] Add tests asserting the behavior of a `zIndexOffset` option
* [x] Add the `zIndexOffset` option, making the test suite :green_heart:
* [x] Add documentation for the `zIndexOffset` option

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -363,3 +363,11 @@ Day of the week start. 0 (Sunday) to 6 (Saturday)
 
 .. figure:: _static/screenshots/option_weekstart.png
     :align: center
+
+
+zIndexOffset
+------------
+
+Integer.  Default: 10
+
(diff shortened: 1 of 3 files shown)


hebbet:
LGTM, merged.

laserlemon:
Does the project have a release schedule so I can be aware of when a new version will ship with this change? Thank you!

hebbet:
@laserlemon see #1497

PR #1508, opened 2015-07-30
Temporary IE fix

divpreet:
This would fix datepicker crashing on January 0100 dates


divpreet:
Adding a temporary fix to avoid crashes on IE and firefox for dates like Jan 0100.

hebbet:
thanks

PR #1516, opened 2015-08-05
Update bootstrap-datepicker.zh-TW.min.js

andy00005555:
add clear zh-tw word


hebbet:
please don't change the dist files. 

edit js/locales.

acrobat:
locale file is already ok for this language. I will re-generate the dists files soon!

PR #1503, opened 2015-07-28
Issue #1489 wrong month or dates shown

throup:
Fix for https://github.com/eternicode/bootstrap-datepicker/issues/1489

PR #1524, opened 2015-08-13
Fixing month in title

HNygard:
Using the viewDate for displaying month in title. In commit 67a2482691d6c253064d44a47f83cde7b21cd6e3, this line was changed to introduce improved translation. The old version used the variable "month" which is based on viewDate. The new version created a new Date at the start of the month.

In my timezone, +2, the result of new Date(2015, 7) is "Sat Aug 01 2015 00:00:00 GMT+0200 (W. Europe Daylight Time)".  DPGlobal.formatDate then uses date.getUTCMonth(), which is now 6 (instead of 7). This again mean that we end up with "July 2015" in the title.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -922,7 +922,7 @@
 			if (isNaN(year) || isNaN(month))
 				return;
 			this.picker.find('.datepicker-days thead .datepicker-switch')
-						.text(DPGlobal.formatDate(new Date(year, month), titleFormat, this.o.language));
+						.text(DPGlobal.formatDate(d, titleFormat, this.o.language));
 			this.picker.find('tfoot .today')
 						.text(todaytxt)
 						.toggle(this.o.todayBtn !== false);


HNygard:
The fix was tested in the project that experienced the problem. Used master branch of https://github.com/hnygard/bootstrap-datepicker/

hemp:
:+1:   Sorry it was all my fault!

Do you think having a unit-test for this specific situation would help?  In hindsight, I wouldn't have made the goof and opened the PR if a test was failing.

acrobat:
Yes please add a test to avoid any future regressions! Thanks!

vsn4ik:
Current `npm test` on branch `master` fails. After this fix - done. Check not needed.

acrobat:
ping @HNygard is it possible to add some tests to avoid any future regressions

HNygard:
Sorry. I don't have the time to create that :-(

vsn4ik:
Issue already fixed by merge #1503. But your code shortest.

About the test. I do not know how all tests in #1503 passed, but now, if change `new UTCDate` to `new Date` 58 test failed.

For this situation, the test will write only if somehow indicate qunit (phantomjs) time zone. I tried to do it for 2 hours.

acrobat:
Closing this PR as a fix was merged with PR #1503 and tests are ready to be merged in PR #1654

Thanks for your time @HNygard!

PR #1527, opened 2015-08-16
Double click the selected date to hide picker

tillson:
It's a bit hacky, but it seems like the only way (correct me if I'm wrong).

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -409,7 +409,7 @@
 				}]);
 			}
 
-			this._secondaryEvents = [
+      this._secondaryEvents = [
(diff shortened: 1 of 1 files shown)


vsn4ik:
Thanks for the PR, but we have to reject this feature.

mikeerickson:
@vsn4ik Since you have rejected this feature, is there a different method for enabling this?  Perhaps a workaround to target it with jQuery?

Azaret:
You can workaround with [`.dbclick()`](https://api.jquery.com/dblclick/) and [`.hide()`](https://bootstrap-datepicker.readthedocs.io/en/stable/methods.html#hide).

PR #1543, opened 2015-08-28
enter when forceParse=false should not parse the date

bertBruynooghe:


bertBruynooghe:
not properly linked to the issue

PR #1546, opened 2015-08-31
Fix date-reset issue when navigating to next with invalid date

cvanleeuwen:
When typing/pasting an invalid formatted date into the textbox (e.g. 2000-1-1 for a picker with mm/dd/yyyy) into a box, the browser will stall when clicking next.

I've traced the issue to the date-resetting loop in the moveMonth.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -49,6 +49,9 @@
 			return this[method].apply(this, arguments);
 		};
 	}
+	function isValidDate(d) {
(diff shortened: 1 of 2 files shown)


cvanleeuwen:
I'm not sure if I followed the PR guidelines correctly (it's my first!). Any suggestions are very much welcome!

robvdlv:
Ah we encountered this very issue last week. Thanks for fixing!

acrobat:
Thanks @cvanleeuwen!

PR #1547, opened 2015-08-31
Chage `fill` method call position

ufocoder:
I want add extra CSS classes to calendar cell with help of callback function of `beforeShowDay`, but there's a wrong method call order. 
Consequently, I suppose plugin must call `fill` method in this place like in other, other words `fill` method must be called only after `setValue` was called.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1248,11 +1248,11 @@
 			if (!which || which  === 'view')
 				this.viewDate = date && new Date(date);
 
-			this.fill();
 			this.setValue();
 			if (!which || which  !== 'view') {
 				this._trigger('changeDate');
 			}
+			this.fill();
 			var element;
 			if (this.isInput){
 				element = this.element;


ufocoder:
#1551

acrobat:
Can you provide a jsfiddle example to show the exact problem with the fill/beforeShowDay?

ufocoder:
URL: https://jsfiddle.net/bxgu1vw8/3/ 
Red background must be always instead default blue background, but if you click to the several dates on calendar you discover that is not.

vsn4ik:
Updated example https://jsfiddle.net/bxgu1vw8/4/. LGTM.

acrobat:
@ufocoder Can you add some tests to check this new behavior? That way we are sure this fix won't be reverted in the future and we are now sure it won't break the datepicker. Thanks!

vsn4ik:
I believe that everything is correct and this behavior is not an error.

Azaret:
Yeah there is not issue for me. `beforeShowDay` is called because the picker is updated after changing the date. So there is nothing wrong here.
What could be do is to add an event on a selected date or tilde with the element as arguments, but that will be in the `2.0`.
By the way, if the aim is to add a class to the selected date, active is already added, so what the hell ?
That's a nay for me.

PR #1556, opened 2015-09-02
Added better year detection

dswitzer:
I refactored the code to do better year detection. There are 2 changes:

* When a user enters a 2 digit year, the code now does a better job of guessing intent. If the 2 digit year is more than 10 years in the future, it'll use the past century instead. So in the 2015, a value between 00-25 would assume the century is 2000. For 26-99, it'll assume the user meant 1900 (e.g. 1/31/10 = Jan 31, 2010, 1/31/26 = Jan 31, 1926). People don't generally enter future dates too far in advance, but it's very common to enter dates from 50+ years ago.
* I changed the "yyyy" parse so that if the year is under 100, it assumes the user meant a 4 digit year. I can't think of a valid use case of when someone would really be entered Jan 1, 0015 as a date, but it's extremely common for users to enter a 2 digit year when a 4 digit year is asked for.

Hopefully you find this changes useful.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1745,14 +1745,28 @@
 			}
 			parts = date && date.match(this.nonpunctuation) || [];
 			date = new Date();
+
+			function adjustYear(year){
+				// if year is 2 digits or less, than the user most likely is trying to get a recent century
+				if( year < 100 ){
+					year += 2000;
+					// if the new year is more than 10 years in advance, use last century
+					if( year > ((new Date()).getFullYear()+10) ){
+						year -= 100;
+					}
+				}
+
+				return year;
+			}
+
 			var parsed = {},
 				setters_order = ['yyyy', 'yy', 'M', 'MM', 'm', 'mm', 'd', 'dd'],
 				setters_map = {
 					yyyy: function(d,v){
-						return d.setUTCFullYear(v);
+						return d.setUTCFullYear(adjustYear(v));
 					},
 					yy: function(d,v){
-						return d.setUTCFullYear(2000+v);
+						return d.setUTCFullYear(adjustYear(v));
 					},
 					m: function(d,v){
 						if (isNaN(d))
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1748,10 +1748,10 @@
 
 			function adjustYear(year){
 				// if year is 2 digits or less, than the user most likely is trying to get a recent century
-				if( year < 100 ){
+				if (year < 100){
 					year += 2000;
 					// if the new year is more than 10 years in advance, use last century
-					if( year > ((new Date()).getFullYear()+10) ){
+					if (year > ((new Date()).getFullYear()+10)){
 						year -= 100;
 					}
 				}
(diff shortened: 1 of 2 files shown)


hebbet:
LGTM,

acrobat:
Looks good! I think we should have some more tests to checking this behavior and have a little piece in the docs of this way of parsing the years

acrobat:
ping @dswitzer

dswitzer:
@acrobat 

Was the comment about more tests for me? If so, that wasn't clear. I thought it was internal dialog for the collaborators.

acrobat:
Ah yes, sorry if it was not clear! Yes we would like some tests to verify all possible year inputs and an little block in the docs about the new way of parsing the years

js/bootstrap-datepicker.js
```diff
@@ -1745,14 +1745,28 @@
...
+
+				return year;
+			}
+
 			var parsed = {},
 				setters_order = ['yyyy', 'yy', 'M', 'MM', 'm', 'mm', 'd', 'dd'],
 				setters_map = {
 					yyyy: function(d,v){
-						return d.setUTCFullYear(v);
+						return d.setUTCFullYear(adjustYear(v));
```
Azaret: I'm against this! :<
If we explicitly asked for 4 digits year, then year 56 is year 56, not 2056.
I could accept this 2 digits format (tho I think this should be discouraged), then only if we have `yy` format we adjust the date.

ajb:
Seems like this should be an opt-in piece of functionality for backwards-compatibility, or at the very least, opt-out so that folks like @Azaret can keep using 1st century dates.

dswitzer:
In my experience, I can ask for a 4 digit year, but people will still enter 2 digit years. I could see having the option for forcing the explicit 4 digit, for backward compatibility. I'm sure there are some use cases where that makes sense (especially if you're doing some kind of historic data entry), but I would assume most times a user enters 2 digits when 4 has been asked for, they're attempting a recent date.

Azaret:
My argument here is not to fit one user case (historical stuff) but more like: As there is, and will be, several use cases where the expected output when typing 2 digits years will be different, then it should not try to assume what the user meant.

In my opinion the picker should be agnostic from assuming what the user meant. Because if we have something agnostic then all use cases will work. If you start to assume when the user meant, then you'll restrict the use cases.

Also you put the limit to your limit to know if the user meant 21's or 20's century at 10 years, but this is totally arbitrary (I kinda don't like arbitrary stuff). And there this some use case where this won't work, for example copyrights; For copyrights it should assume when the user type 20 that he meant 2020, even 2120. But I'll agree that this is a specific use case, but so it is to take this 10 years limit.

And in the case the user really meant to enter a date under 100 for any reason, then like the core `Date` object you'll introduce the Y2K bug and won't let the user input those years, so at least it should be an optional feature.

At the end, it's just a personal belief that assuming is bad but I can understand it. Just it should be at least optional, we should be able to set any year.

ajb:
The way I see it, this behavior makes a lot of sense for some, and would not be appropriate for others.

I've opened PR https://github.com/eternicode/bootstrap-datepicker/pull/1631 with a modified version that lives behind a disabled-by-default option.

I'm happy to run through the use-case of why this would make sense, in case that helps getting it merged.

vsn4ik:
:-1: for this.

ajb:
Folks who :-1:'d this: please let me know what you think of https://github.com/eternicode/bootstrap-datepicker/pull/1631.

acrobat:
Thanks for the PR @dswitzer but I'm closing this as #1631 got merged and is a better way of solving this problem!

PR #1558, opened 2015-09-04
added keepEmptyValues option

### LINKED ISSUE #1454: Daterange. Disable automatic setting of other empty date field?
I'm using two datepickers that can be used to set a range, this works great. The user can also submit just a start or end date by leaving one field empty. 

Currently a daterange picker always sets another empty field to the same value the user just entered. Is there a way to disable this? Would there be interest in an option to disable this? I could try to make a pull request.

cheers

steffendietz:
If this option is set to true, it will stop propagation of a selected
date to other, currently empty, pickers in a date range setup.

Default for this option is false, to keep the current behavior.

fixes #1454

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -284,6 +284,14 @@ Boolean. Default: false
 If true, selecting a year or month in the datepicker will update the input value immediately. Otherwise, only selecting a day of the month will update the input value immediately.
 
 
+keepEmptyValues
+---------------
+
+Boolean. Default: false
+
+Only effective in a range picker. If true, the selected value does not get propagated to other, currently empty, pickers in the range.
+
+
 inputs
 ------
 
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1581,6 +1581,9 @@
 		});
 		delete options.inputs;
 
+		this.keepEmptyValues = options.keepEmptyValues || false;
+		delete options.keepEmptyValues;
+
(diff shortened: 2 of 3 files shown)


vsn4ik:
@steffendietz Great! Can you rebase the branch agianst the current master?

steffendietz:
done 🚒

hebbet:
Thanks, LGTM. :+1:

docs/options.rst
```diff
@@ -284,6 +284,14 @@ Boolean. Default: false
 If true, selecting a year or month in the datepicker will update the input value immediately. Otherwise, only selecting a day of the month will update the input value immediately.
 
 
+keepEmptyValues
+---------------
+
+Boolean. Default: false
```
acrobat: Can you add the option an its default value to the overview at the bottom of this docs file

js/bootstrap-datepicker.js
```diff
@@ -1581,6 +1581,9 @@
 		});
 		delete options.inputs;
 
+		this.keepEmptyValues = options.keepEmptyValues || false;
```
acrobat: The default value should be adde to the defaults array, so we can keep all default values in one place

acrobat:
@steffendietz The PR was merged butI had some extra comments, can you fix those in a separate pr against master? Thanks!

steffendietz:
@acrobat Yes. Will do so tomorrow.

Issue #1567, opened 2015-09-10
Wrong algorithm in merged PR


vsn4ik:
PR #1329. Commit b65143565b88cf4cb749c3d11f5f11509f5987eb.

`yorient` should be exhibited only when it is set `'auto'`.

Also `top` position wrong after merge. See: http://jsbin.com/pofemazoha/edit?html,js,output.

Issue #1529, opened 2015-08-17
Datepicker orientation failer


rammi987:
Hello, nice picker
small issue with it, when i click on the input field to show up the calender, the orientation fail.

The calender show on the top of the input witch is fair enuge but the poniting arrow points up away from the input field, do i try to force it to be "bottom lef", issues will be the pointing arrow will point down agian away from the field?

<code>
export function init(scope) {
	$(scope).datepicker({
		format: 'dd/mm-yyyy',
		todayHighlight: true,
		todayBtn: "linked",
		 orientation: "top left",
	});
}
</code>

<code>
input.datepicker(type="text")
</code>

jcdavison:
Hey @rammi987 did you find a solution to this? I'm having the same issue.

https://www.dropbox.com/s/418fyc1imcy67qi/Screenshot%202015-08-25%2012.09.40.png?dl=0

MartijnR:
same issue

jcdavison:
I actually noticed that using the rails gem associated with adding this js to the asset pipeline was having issues, I pulled the js manually into the app I was working and the situation resolved itself.  The likely culprit, in my case, was https://github.com/Nerian/bootstrap-datepicker-rails not the source js itself. probably working off and old version of the js.

MartijnR:
Thanks @jcdavison! Turns out I was also using an old js version.

rammi987:
i self just downgrate bootestrap version and it works agian

acrobat:
So whith which version of the datepicker do you guys have this bug? master(1.5) or a specific lower version?

MartijnR:
I use 1.4.0 (the latest version published on npm), but in my case it turned out to not be a bug.

hebbet:
fixed on master with f08ad04.

Issue #1496, opened 2015-07-21
Datepicker in modal breaks layout


tfr:
Hi,

I have a user interface which utilizes Bootstrap modal "popups". It's built so that in case the content of the popup is taller than the user's screen, it will extend the modal to the height of the screen and then make the content scrollable.

This works fine, except that if the content contains a datepicker, then, upon opening datepicker, it breaks badly - it breaks the scrolling and extends the modal content to its original length (through top and bottom of the screen, which, though, remains unscrollable, and also outside the modal window horizontally.

Sample:

Before clicking the datepicker (on "Purchase date"):
https://www.dropbox.com/s/vju3ppooqvushch/Screenshot%202015-07-21%2012.48.10.png?dl=0

After:
https://www.dropbox.com/s/6vujribqcl3en6k/Screenshot%202015-07-21%2012.48.15.png?dl=0

Any idea that might be causing this behaviour?

Bootstrap v3.3.0
jQuery v1.11.1
Bootstrap-datepicker latest and greatest as of today (1.5.0-dev according to package.json)

TIA,
Indrek

hebbet:
the current dist folder in master is not up2date.
please try running grunt or use 1.4

tfr:
I was on 1.4 before. Tried upgrading to 1.5.0-dev to see if it has been fixed.

tfr:
Here's a test case that reproduces this bug: http://datepickerbug.codelab.solutions

tfr:
Ah, one more thing - if you open the modal and close it immediately (without actually opening the datepicker), the datepicker will be shown after the modal has closed.

tfr:
Ah, one more thing - if you open the modal and close it using the escape key (not the "Cancel" button at the bottom), the datepicker will be shown after the modal has closed.

hebbet:
seems to be fixed with f08ad04, closing.

hebbet:
@tfr can you test the current master?

tfr:
Updated http://datepickerbug.codelab.solutions <http://datepickerbug.codelab.solutions/> - it seems slightly better in some cases, but definitely not fixed.

-- 
Indrek

vsn4ik:
@ttr your link contains not containsactual bootstrap-datepicker (dist). Please update and re test.

tfr:
What do you mean exactly? It contains the latest master (v1.5.0-dev) that habbet wanted me to test.

vsn4ik:
[...]

Issue #1480, opened 2015-07-13
Incorrect 'auto' orientation


meetraz:
I have an application where the date field is located near the top of the page.  When I click on the field, the date picker is appearing on the top, so half of it is off the page and not visible.  I tried to add data-date-orientation="bottom" to the input attributes, but it's had no effect.   Is there a way to fix this?

tomvo:
I have the same problem and not really have a solution. When I use `orientation: 'bottom right'` (or all the other combinations) it seems only to adjust the horizontal position, not the vertical one.

tomvo:
oh jeez, I just figured it out, see this one: https://github.com/eternicode/bootstrap-datepicker/issues/1406
It's just mixed-up.

hebbet:
fixed on master with f08ad04

PR #1571, opened 2015-09-13
Fix vertical orientation

### LINKED ISSUE #1286: Grunt screenshot doesn't show datepicker correct
We running the phantomjs screenshots grunt task, the screenshots don't get generated correctly

The datepicker popup show above the field instead of under the field, that way the popup isn't shown on the screenshots

But I don't know how to debug it or why it is broken (because running a manual example the popup shows correct). So anyone know how to debug/fix this?

Just run following commands to reproduce (compare with the images currently in the docs/_static/screenshots dir

```shell
$ npm install
$ grunt screenshots
```

This is the last issue before the 1.4.0 release is ready!

/cc @eternicode @hebbet @abeverley

### LINKED ISSUE #1517: Issue with calendar appearing off top screen
I have created a very simple demo app trying to get the bootstrap-datepicker to work.   

When I click in the area that has the datepicker, the calendar appears off top part of screen.  

The only way it looks right is if I have a lot of stuff above it to force the control down on page (and then it barely makes it).   From what I am observing, it always opens on top of input field (never below which I would think it should do).

Other than the visual issue, everything else works as expected.

Here the files I am including (same issue occurs when i am not using the parallel time popup)

<pre>

    <script src="lib/jquery.timepicker.min.js"></script>
    <link href="lib/jquery.timepicker.css" rel="stylesheet">

    <script src="lib/bootstrap-datepicker.js"></script>
    <link href="lib/bootstrap-datepicker3.css" rel="stylesheet">

</pre>

vsn4ik:
Fix #1286, #1567, #1529, #1480.
Maybe fix #1517, #1496, #1364, #1146, #1088, #1035.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -678,7 +678,6 @@
 				visualPadding = 10,
 				container = $(this.o.container),
 				windowWidth = container.width(),
-				windowHeight = container.height(),
 				scrollTop = container.scrollTop(),
 				appendOffset = container.offset();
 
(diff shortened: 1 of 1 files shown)


hebbet:
LGTM, will do some tests later today for the maybe fixes issues.

acrobat:
Fix does indeed look good!

hebbet:
merged, thanks.

PR #1582, opened 2015-09-18
Added custom formatting support

### LINKED ISSUE #930: ISO date formatting
Hey there, excellent plugin, waiting patiently for some resolution on an integrated timepicker, however currently I am in need to pick ISO dates from the calendar, which I temporary achieve by allowing the user to only select mondays, however the result value is not displayed as I wish and can not be interpreted by the plugin.

PHP's [date](http://php.net/manual/en/function.date.php) supports it via `o-W` format, so does the dreadful jQuery UI: http://api.jqueryui.com/datepicker/#utility-formatDate

Could you please add similar functionality?

abibell:
Datepicker options now supports custom formatting functions
```javascript
format: {
	parse: function(date, format, language) { /* Your custom parse to Date code goes here */ },
	format: function(date, format, language) { /* Your custom format to string code goes here */ }
} 
```

Solves https://github.com/eternicode/bootstrap-datepicker/issues/930


vsn4ik:
:+1: for merge.

acrobat:
PR looks good! But I'm thinking about the option names, I'm not sure but maybe something like `toDisplay` and `toValue` are more clear names?

abibell:
Absolutely @acrobat, they are more functional names rather than Tech names. I renamed it now.

On a side note should the existing DPGlobal's parseDate() and formatDate() function names be upgraded too?

docs/options.rst
```diff
@@ -189,6 +189,36 @@ The date format, combination of d, dd, D, DD, m, mm, M, MM, yy, yyyy.
 * M, MM: Abbreviated and full month names, respectively.  Eg, Jan, January
 * yy, yyyy: 2- and 4-digit years, respectively.  Eg, 12, 2012.
 
+Object.
+
+Custom formatting options
```
acrobat: Add an empty line between this text and the bullets. Then the bullets will be rendered correctly!

acrobat:
Just one last comment! Then the PR should be ready to merge!

vsn4ik:
Maybe `if (format.to*)` -> `if (typeof format.to* == 'function')`

acrobat:
That would indeed be a more secure way to check if you can use the formatter functions

abibell:
The suggestions have been added @acrobat and @vsn4ik. Good to go.

acrobat:
Thanks @abibell!

abibell:
Should that close https://github.com/eternicode/bootstrap-datepicker/issues/930 @acrobat?

acrobat:
Yes indeed, closed the issue!

Issue #1559, opened 2015-09-04
update() method doesn't apply to all datepickers


Dimas-sc:
I've updated the Datepicker to the last version and now when I do:

$('#modal-planificacio .input-group.date').datepicker('update');

It only updates the datepicker of the first element (i've two datepickers). Before it updates all of them. Now this works, but isn't a good solution:

$('#modal-planificacio .input-group.date').first().datepicker('update');
$('#modal-planificacio .input-group.date').last().datepicker('update');

acrobat:
Possibly related to #1572

acrobat:
PR #1586 fixed this issue, please test against the latest master!

PR #1586, opened 2015-09-19
Fix jQuery chain breaks

### LINKED ISSUE #1572: For some methods chain breaks on the first jQuery element
`$('input').datepicker('show')` if `$('input').length > 1` methods chain breaks on the first jQuery element.
For reproduce run `grunt screeshots` (see `docs/_screenshot/script/screenshot.js`) and see generated images.

It was created by 847ac88bee7709dcfd12927cd851cc339f1442dd
```js
...
if (internal_return !== undefined)
  return false;
...
```

Maybe delete `return this;` in `show`, `remove`, `hide`, and `update` methods or change to `return;`.

vsn4ik:
Fix #1572, #1559.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1612,14 +1612,20 @@
 			}
 			if (typeof option === 'string' && typeof data[option] === 'function'){
 				internal_return = data[option].apply(data, args);
-				if (internal_return !== undefined)
-					return false;
 			}
 		});
-		if (internal_return !== undefined)
-			return internal_return;
-		else
+
+		if (
+			internal_return === undefined ||
+			internal_return instanceof Datepicker ||
+			internal_return instanceof DateRangePicker
+		)
 			return this;
+
+		if (this.length > 1)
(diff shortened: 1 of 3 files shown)


hebbet:
LGTM.

acrobat:
Thanks for the PR @vsn4ik, I'm going to test the patch but looks indeed good!

acrobat:
@vsn4ik this change breaks something, run `grunt screenshots` locally and check the `demo_head.png` screenshot, the datepickers are now printer under each other. Without the change it prints "ok" (only the first datepicker, but that's what this issue is about)

Can you check what is breaking this?

vsn4ik:
@acrobat, i do not see the same problem in `*.png`
Try changing the `docs/_screenshots/demo_head.html` directly `.row` to `.row-fluid`.

acrobat:
Hmm yes indeed that fixes the problem! I will commit that to master and then I will merge the PR! Again one step closer to the 1.5 release, thanks!

PR #1589, opened 2015-09-20
Some fixes in docs

vsn4ik:
Change range example BS2 to BS3.

### DIFF
--- a/docs/markup.rst
+++ b/docs/markup.rst
@@ -24,8 +24,10 @@ Adding the ``date`` class to an ``input-group`` bootstrap component will allow t
 .. code-block:: html
 
     <div class="input-group date">
-        <input type="text" value="12-02-2012">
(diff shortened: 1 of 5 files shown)


docs/markup.rst
```diff
@@ -73,11 +77,13 @@ Example to save the embedded datepicker value to a hidden field
 .. code-block:: html
 
     <div id="datepicker" data-date="12/03/2012"></div>
-    <input type="hidden" id="my_hidden_input" />
+    <input type="hidden" id="my_hidden_input">
```
acrobat: the end of the input tag should be />

acrobat:
only one remark, the rest of the PR looks good!

docs/markup.rst
vsn4ik: Not for `input`. http://www.w3.org/TR/html5/syntax.html#void-elements

acrobat: Ok!

acrobat:
Thanks @vsn4ik!

PR #1591, opened 2015-09-21
Fix an issue with clicking outside a span, td, or th

czert:
The click handler will search for the closest ancestor that is a span, td, or th. This will sometimes find an element outside the picker itself and treat it as a datepicker control, which will result in setting the year to zero.

To see this bug in action, wrap the picker in a span, set a startDate, and in the "month" view, click on the place where a "<<" button would normaly be.

This fix adds a simple check that the target is actually inside the datepicker.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1117,7 +1117,7 @@
 			e.stopPropagation();
 			var target = $(e.target).closest('span, td, th'),
 				year, month, day;
-			if (target.length === 1){
+			if (target.length === 1 && $.contains(this.picker.get(0), target.get(0))){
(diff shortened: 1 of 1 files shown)


acrobat:
@czert thanks for this PR! Can you provide a jsfiddle to display the problem?

Azaret:
Can we test this with #1624 ?
The `click` function has been rewrote.

czert:
Hi, sorry, can't really afford to spend any time on this now. Maybe someone else can provide the fiddle based on my description?

> To see this bug in action, wrap the picker in a span, set a startDate, and in the "month" view, click on the place where a "<<" button would normaly be.

Issue #616, opened 2013-08-28
Trigger an event in the fill-function


terbra:
It would be nice to have an event triggered at the end of the fill function to be notified when the rendering is complete.

Azaret:
Which event is aimed exactly ? because in few case `fill()` is not the last event called.

acrobat:
Some sort of aferRender event where you can hook into

PR #1605, opened 2015-10-07
[Feature] New event afterFill

Efiware:
I have added the event afterFill that (surprise) gets triggered when the fill method has been executed.
By having this event, it will be possible to modify the layout after rendering. This is nice when you want to highlight a month or week when using Datepicker embedded.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1070,6 +1070,7 @@
 				year += 1;
 			}
 			yearCont.html(html);
+			this._trigger('afterFill');
 		},
 
 		updateNavArrows: function(){


acrobat:
Maybe we should call this event something more like "afterRender" ?
/cc: Possible fix for #616

Efiware:
Sure whatever

vsn4ik:
@Efiware `afterChange` or `afterRender`?

Efiware:
Sorry - was a bit fast there :)
Updated to afterRender as requested.

acrobat:
Great thanks! I will test this patch tonight and if it works I will merge this for 1.5.0 which should be released this weekend!

Efiware:
:thumbsup:

acrobat:
The event triggers correctly when you select a new date, but on load the event is not triggered. My idea is that it should trigger also then, don't know if i'm correct about that

acrobat:
The event triggers correctly when you select a new date, but on load the event is not triggered. My idea is that it should trigger also then, don't know if i'm correct about that. What do you think?

/cc: @hebbet, @vsn4ik

Azaret:
I believe it's because `fill` is only called first when you focus the input, not when it instantiate.
But I guess it's not an issue, it should rather be another event triggering when loading ends.
It wont make sense to trigger it at load since nothing if rendered yet.

PR #1610, opened 2015-10-11
Fixed 'tip' triangle for top-positioned calendars

jesperronn:
Fix for the orientation-top class (used when calendars renders above the input field).

The problem was that the 'tip' triangle was left dangling above the calendar, and it should be moved to the bottom.

The fix is really simple: when we apply a position-bottom value, then we MUST reset the top value. I do that with `top: initial`.

Browser: Chrome Version 47.0.2526.8 dev (64-bit)

### DIFF
--- a/less/datepicker3.less
+++ b/less/datepicker3.less
@@ -43,11 +43,13 @@
 			bottom: -7px;
 			border-bottom: 0;
 			border-top:    7px solid @dropdown-border;
+			top: initial;
 		}
 		&.datepicker-orient-top:after {
 			bottom: -6px;
 			border-bottom: 0;
 			border-top:    6px solid @dropdown-bg;
+			top: initial;
 		}
 	}
 	> div {


acrobat:
Can you provide some screenshots or jsfiddle example to show before and after fix?

acrobat:
Ping @jesperronn

jesperronn:
@acrobat sorry to have kept you waiting. I am now on to other tasks and am not currently able to provide screenshots before and after.

I originally fixed it in a private customer repository with the following text:

> Fixed 'tip' triangle for top-positioned calendars #1610
> 
> Fix for the orientation-top class (used when calendars renders above the input field).
> 
> The problem was that the 'tip' triangle was left dangling above the calendar, and it should be moved to the bottom.
> 
> The fix is really simple: when we apply a position-bottom value, then we MUST reset the top value. I do that with `top: initial`.
> 
> Browser: Chrome Version 47.0.2526.8 dev (64-bit)

I understand it must be frustrating to have pull requests left open and dangling. I guess this leaves you with two options:

1. close the PR unmerged (as I can currently not provide screenshots)
2. merge the PR in case you are confident it will not introduce other issues.

It's your call. Sorry about the missing screenshots

acrobat:
No problem @jesperronn! I will review the pr locally first and see the before and after case that way!

jesperronn:
Closing this as it has been open for ages. Feel free to reopen if you feel this is still relevant.

PR #1437, opened 2015-06-18
Picking a decade #753

Azaret:
This is a first draft, I won't be against someone else running intensive test too.

- [x] Decade picker
- [x] Century picker
- [x] Qunit tests
- [x] Docs

Issue #753, opened 2013-12-19
picking a decade


jogaco:
Any hint on how could I implement a decade picker?, like making the decade title on the year picker selectable? I could give it a try myself.

Azaret:
While doing this, I though we could go a step further, with picking a century.
There still at the start of the 21th one and many people are born in the 20th.

acrobat:
Fixed in #1623

PR #1623, opened 2015-10-17
PIcking a decade and a century - Rebase of #1437

Azaret:
So, Add the decade picker and century pickers.
This is a rebase of #1437 based on #753.


Azaret:
Or, we might need to split up the fill function.

vsn4ik:
LGTM.

acrobat:
Thanks @Azaret!

Issue #1625, opened 2015-10-17
Datepicker title not centered when displaying calendar weeks


mologie:
Hello,

I found that in the current release candidate (1.5 RC1), with the following configuration, the title is not properly centered:

```js
$(document.forms.addScSollForm.sollDate).datepicker({
	calendarWeeks: true,
	todayHighlight: true,
	daysOfWeekDisabled: [0,2,3,4,5,6],
	title: "Select calendar week"
});
```

The reason appears to be that the generated title `th` element has a hardcoded colspan of 7 columns. However, when displaying calendar weeks, there are 8 columns.

A screenshot of the issue is attached.

![capture](https://cloud.githubusercontent.com/assets/597682/10559661/c3c84626-74f6-11e5-8511-73d9afbc17de.PNG)

vsn4ik:
Fixed, but not closed...

/cc @acrobat, @hebbet.

hebbet:
thanks ;-)

PR #1626, opened 2015-10-17
Fix datepicker title not centered when displaying calendar weeks

mologie:
This pull requests fixes bug #1625. The fix is trivial and makes use of an existing feature; the potential for a regression is minimal.

Cheers, Oliver


vsn4ik:
LGTM

acrobat:
Thanks @mologie!

PR #1631, opened 2015-10-20
Add "assumeNearbyYear" option

ajb:
Rebase / modification of #1556. My hope is that this can be merged.

If the user opts-in by setting `options.assumeNearbyYear` = true:

- When a user enters a 2 digit year, the code now does a better job of
guessing intent. If the 2 digit year is more than 10 years in the future,
it'll use the past century instead. So in the 2015, a value between 00-25
would assume the century is 2000. For 26-99, it'll assume the user meant
1900 (e.g. 1/31/10 = Jan 31, 2010, 1/31/26 = Jan 31, 1926). People don't
generally enter future dates too far in advance, but it's very common
to enter dates from 50+ years ago.

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -176,6 +176,17 @@ Boolean.  Default: true
 Whether or not to force parsing of the input value when the picker is closed.  That is, when an invalid date is left in the input field by the user, the picker will forcibly parse that value, and set the input's value to the new, valid date, conforming to the given `format`.
 
 
+assumeNearbyYear
+
+----------
+
+Boolean or Integer.  Default: false
+
+If true, manually-entered dates with two-digit years, such as "5/1/15", will be parsed as "2015", not "15". If the year is less than 10 years in advance, the picker will use the current century, otherwise, it will use the previous one. For example "5/1/15" would parse to May 1st, 2015, but "5/1/97" would be May 1st, 1997.
+
+To configure the number of years in advance that the picker will still use the current century, use an Integer instead of the Boolean true. E.g. "assumeNearbyYear: 20"
+
+
 format
 ------
 
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -236,7 +236,7 @@
 					if (o.startDate instanceof Date)
 						o.startDate = this._local_to_utc(this._zero_time(o.startDate));
 					else
-						o.startDate = DPGlobal.parseDate(o.startDate, format, o.language);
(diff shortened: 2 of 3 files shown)


dswitzer:
Would it make more sense for options.assumeNearbyYear to allow "false" to skip the behavior (mimicking the current functionality) or an integer which is the threshold for the assumption. So,

options.assumeNearbyYear = false; // exact value used
options.assumeNearbyYear = 10; // if date is more than 10 year in future, then past century is used
options.assumeNearbyYear = 25; // if date is more than 25 year in future, then past century is used

That might provide more flexibility. We could also make the length a separate option.

ajb:
Thanks Dan, great suggestion. If folks want it, I think a separate option
makes the most sense.

On Thu, Oct 22, 2015 at 11:57 AM, Dan G. Switzer, II <
notifications@github.com> wrote:

> Would it make more sense for options.assumeNearbyYear to allow "false" to
> skip the behavior (mimicking the current functionality) or an integer which
> is the threshold for the assumption. So,
>
> options.assumeNearbyYear = false; // exact value used
> options.assumeNearbyYear = 10; // if date is more than 10 year in future,
> then past century is used
> options.assumeNearbyYear = 25; // if date is more than 25 year in future,
> then past century is used
>
> That might provide more flexibility. We could also make the length a
> separate option.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/eternicode/bootstrap-datepicker/pull/1631#issuecomment-150270319>
> .
>



-- 
Adam Becker
(951) 9-BECKER
@AdamJacobBecker

Azaret:
That's actually more thoughtful and customization.
I think one option is fine, having two won't benefit much, since when `assumeNearbyYear` will be `false` the second option won't be used, and when `true` the second or its default will be, perforce, used.

ajb:
I don't feel too strongly about this, but I'd rather the user not have to know the "magic" number of "10" in order to opt-in to this functionality.

E.g. I prefer `assumeNearbyYear: true` to `assumeNearbyYear: 10`.

Azaret:
It could accept `assumeNearbyYear: true` which will set the default value at 10

ajb:
Great idea -- I just made the modification.

Azaret:
If I may, don't reformat the code please, it make harder to read the diff.

ajb:
Didn't notice that. The problem is that there is an `.editorconfig` file with the wrong values. (https://github.com/eternicode/bootstrap-datepicker/issues/1636)

acrobat:
Thanks @ajb!

Issue #1628, opened 2015-10-19
'change' event at initialization - Maximum call stack size exceeded


floriancargoet:
Fails with 1.5.0
Works with 1.4.1

**Summary**: 
When a 'change' handler is set up before instantiation, a call to $().datapicker() in the handler triggers an infinite recursive call ending with an exception: Uncaught RangeError: Maximum call stack size exceeded.

**Minimal code to reproduce**:
``` html
<input type="text" value="02-16-2012" data-provide="datepicker">
```
```javascript
var $el = $("input");

$el.on("change", function () {
  // maximum call stack since the datepicker is being instantiated
  $el.datepicker("getUTCDate");
});
```
[Codepen](http://codepen.io/floriancargoet/pen/JYOVmw)

Click on the input, wait for the exception.

**Details**:
The root cause is that the datepicker triggers a 'change' event during instantiation (introduced in https://github.com/eternicode/bootstrap-datepicker/commit/99c6e40a8d8a6047b458722cbdd3f7b2cb3a8da5). 
The 'change' event calls the handler which calls $().datepicker(). Since the instantiation is not finished, this call creates a new instance. Infinite recursive call.

Azaret:
this, [line 1588](https://github.com/eternicode/bootstrap-datepicker/blob/master/js/bootstrap-datepicker.js#L1588)
```js
$this.data('datepicker', (data = new Datepicker(this, opts)));
```
in the instance function should not be done like this.
Data should rather be set in the constructor, at the beginning of the creation of the instance.
I can fix this later today I guess.

Azaret:
That looks nice: http://codepen.io/anon/pen/OyzKGy
Gonna PR it.

floriancargoet:
Can't load your script on codepen.
```
GET https://bsdp.wesiin.com/js/bootstrap-datepicker.js net::ERR_INSECURE_RESPONSE
```

Azaret:
I guess you can directly open the script to accept the cert. My cert was badly made, i have to regenerate it. Or maybe codepen accept none https resources, I used https because jsfiddle only accept https.

Azaret:
There http://codepen.io/anon/pen/pjajwB I changed the files with the one from the PR

floriancargoet:
I confirm the issue is fixed on your codepen.

acrobat:
Fixed in #1633

PR #1633, opened 2015-10-21
Fixing looping when calling dp inside change event #1628

Azaret:
This should solve the issue reported on #1628.


hebbet:
LGTM.

acrobat:
Thanks @Azaret!

PR #1637, opened 2015-10-22
Add CSS sourcemap

vsn4ik:


acrobat:
@vsn4ik When I generate the css files (from less) the resulted files don't contain the top banner anymore, can you please check/fix that?

acrobat:
Thanks @vsn4ik!

PR #1646, opened 2015-10-27
Do not modify viewDate when some methods are called

bielfrontera:
setStartDate, setEndDate, setDaysOfWeekDisabled, setDaysOfWeekHighlighted and setDatesDisabled methods call update to repaint the component.

This causes a troublesome modification of viewDate, and therefore a possible unnecessary change of the displayed month. A new control variable has been added and now, viewDate is not modified when the methods above are called.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -639,34 +639,44 @@
 
 		setStartDate: function(startDate){
 			this._process_options({startDate: startDate});
+			this._allow_update_viewdate = false;
 			this.update();
+			this._allow_update_viewdate = true;
 			this.updateNavArrows();
 			return this;
 		},
 
 		setEndDate: function(endDate){
 			this._process_options({endDate: endDate});
+			this._allow_update_viewdate = false;
 			this.update();
+			this._allow_update_viewdate = true;
 			this.updateNavArrows();
 			return this;
 		},
 
 		setDaysOfWeekDisabled: function(daysOfWeekDisabled){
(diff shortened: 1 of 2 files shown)


bielfrontera:
I detected this issue when I was trying to modify the list disabled dates after `changeMonth` event is fired.

I call a service that provides the list of available dates for each month, and when I try to modify the calendar with this information (after changing the month), the widget returns to the previous month.

Azaret:
I'm not fan of this change actually, feel more like a fix in order to avoid the issue, rather fixing it really.
I'll look into this issue (if you can link a fiddle it would make my life easier), a fix might be to simply initiate viewDate outside of update, or maybe the initiation is bad.

bielfrontera:
Hi, here you have a fiddle: http://jsfiddle.net/bielfrontera/grea506a/ 
My first approach was to avoid calling the update method and call fill method instead. It works for me, but there were two tests that failed, because they expect that date and viewDate are between startDate and endDate after calling setStartDate or setEndDate.

acrobat:
@bielfrontera Can you rebase the PR against the current master?

acrobat:
@vsn4ik @Azaret do you guys see any problems with this change at it could be merged together with #2054

bielfrontera:
Hi @acrobat, I've rebased the PR! Hope it can be merged! :)

guillaumeNeveux:
@acrobat you can merge please ? i need it

guillaumeNeveux (approved):

guillaumeNeveux (approved):

Couss (approved):

PR #1617, opened 2015-10-14
Add scrollTop to position if container is not body

thetimbanks:
This goes with issue #1616 

Original problem: http://jsfiddle.net/thetimbanks/L24bnfpv/
Fixed: http://jsfiddle.net/thetimbanks/L24bnfpv/1/

PR #1647, opened 2015-10-28
Use document scrollTop instead of body scrollTop when using the body container

gbedardsice:
Not sure if this has been discussed before, but couldn't find anything when searching. Apologies if I missed it.

Using `$('body').scrollTop()` yields different results in Chrome / FF [as you can see here](http://stackoverflow.com/questions/12941150/scrolltop-returns-0-in-firefox-but-not-in-chrome). The method always returns `0` on FF, which messes up the automatic placement calculation. Using `$(document).scrollTop()` solves the issue.

Thanks!

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -678,7 +678,7 @@
 				visualPadding = 10,
 				container = $(this.o.container),
 				windowWidth = container.width(),
-				scrollTop = container.scrollTop(),
+				scrollTop = this.o.container === 'body' ? $(document).scrollTop() : container.scrollTop(),
 				appendOffset = container.offset();
 
 			var parentsZindex = [];


acrobat:
@gbedardsice can you verifiy if you still have this issue on the 1.5 branch? As #1617 got merged and looks like ti's fixing something similar!

gbedardsice:
@acrobat, just checked, looks like I still have the issue. You can reproduce the problem here:
http://jsfiddle.net/gbsice/pqrowju2/
bootstrap-datepicker-auto-placement-fail-ff
Looks like this after the fix:
http://jsfiddle.net/gbsice/05xz8dmn/
bootstrap-datepicker-auto-placement-success-ff

acrobat:
Ok, thanks for the feedback @gbedardsice! Will merge this tomorrow! Btw how do you create those gifs?

gbedardsice:
Thanks! I use [LICEcap](http://www.cockos.com/licecap/) for recording my gifs, but I've heard good things about [GifGrabber](http://www.gifgrabber.com).

acrobat:
Thanks @gbedardsice!

gbedardsice:
Thanks for this!

PR #1653, opened 2015-11-03
Changes nonpunctuation to accept unicode characters

hyderali:
Rejecting Unicode characters in `nonpunctuation` is added in #294 to support yyyy年mm月dd format. However rejecting all Unicode in the range \u3400 to \u9fff rejects names of Chinese months too. And calling parseDate with Chinese date having the month format as  MM results in today's date always irrespective of the date passed. Hence don't reject all the Unicode and reject only the three Unicode in Japanese date format.


acrobat:
@hyderali can you provide some unittests for this change? (so we don't break it in the future)

acrobat:
ping @hyderali

hyderali:
@acrobat, sorry for the delay.
I've added the tests in `methods.js` file. Can the tests be in that file or should they be added in a new file ?

acrobat:
Yes the methods suit is good for now! Thanks @hyderali!

PR #1654, opened 2015-11-04
Rebase PR 1524

vsn4ik:
Rebase #1524.
Unfortunately for UTC+0 this test is always performed.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -919,8 +919,8 @@
 				tooltip;
 			if (isNaN(year) || isNaN(month))
 				return;
-			this.picker.find('.datepicker-days thead .datepicker-switch')
-						.text(DPGlobal.formatDate(new UTCDate(year, month), titleFormat, this.o.language));
+			this.picker.find('.datepicker-days .datepicker-switch')
+						.text(DPGlobal.formatDate(d, titleFormat, this.o.language));
 			this.picker.find('tfoot .today')
 						.text(todaytxt)
 						.toggle(this.o.todayBtn !== false);
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -54,7 +54,8 @@ module.exports = function(grunt){
             }
         },
         qunit: {
-            all: 'tests/tests.html'
+            main: 'tests/tests.html',
+            timezone: 'tests/timezone.html'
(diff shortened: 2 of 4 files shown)


acrobat:
Hmm so we need to find a way to have the tests run with different timezone offsets..

Azaret:
If we don't want to implement timepicker until way later, we should get rid of timezone altogether.
Implying get rid of Date object.

I'm not against one day having time support. But in the short term, I think it could ease our development, so we can move on in many issues, focus on other features, and come back on time support later.

acrobat:
Yes time support is something for atleast version 2.0 but could be moved back even further. Maybe we should indeed focus now on making the datepicker not depend on timezone etc.

vsn4ik:
I try to play with `process.env.TZ`

Azaret:
@acrobat Well, for sure we need to not have `new Date()` everywhere at least. Then doing a custom date object is not that hard.
Well to be honest I'm not sure if doing our own object will be a good idea for sure, all I'm sure of is that we are having a lot of troubles by using Date. It feels like a black box that can do unexpected things and we can't control very well.

vsn4ik:
PR updated.

acrobat:
Oh great so now we can run 2 test suits? The normal `test` and `qunit:timezone`?

vsn4ik:
Not, only `test`.

acrobat:
Oh great, missed that! I will do final review soon!

js/bootstrap-datepicker.js
```diff
@@ -919,8 +919,8 @@
 				tooltip;
 			if (isNaN(year) || isNaN(month))
 				return;
-			this.picker.find('.datepicker-days thead .datepicker-switch')
```
acrobat: Why is the thead class removed here?

acrobat: ping @vsn4ik

vsn4ik: My version is more concise. Which option is faster, I do not know. I go back as it was.

acrobat: Oh ok, if you want you can create a separate pr for that change! Just so we don't mix different change sets

acrobat:
Thanks @vsn4ik!

PR #1671, opened 2015-11-13
Data-API support for component, disabled and readonly

fabdouglas:
Add Data-API support for component. Currently, Data-API covers only the input itself, but not the add-on/input-group-addon.
This PR detect the addon, and listen the click on it, exactly as the input, without any JS to be written.

In addition, disable and readonly states are managed.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -2006,9 +2006,13 @@
 
 	$(document).on(
 		'focus.datepicker.data-api click.datepicker.data-api',
-		'[data-provide="datepicker"]',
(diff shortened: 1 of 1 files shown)


acrobat:
@fabdouglas please check the failing test suite! Do we need to add some extra tests to check this behavior?

fabdouglas:
In fact the expected feature is already implemented in a different way but not clearly documented.
My PR adds support of `data-provide="datepicker"` only on `<input>` to detect the right addon, without adding `date' on the container.
But in the failed test I've just seen `data-provide="datepicker"` could be added on `input-group`/`input-append`, and works well.

I've added a PR for the doc in #1674

PR #1672, opened 2015-11-13
Immediate cursor behavior for data-api on component

fabdouglas:
Immediate cursor behavior for data-api on component instead of waiting for the first initialisation for datepicker.

PR #1673, opened 2015-11-13
Immediate cursor behavior for data-api on component

fabdouglas:
Immediate cursor behavior for data-api on component instead of waiting for the first initialisation for datepicker.


acrobat:
See [comment in #1672](https://github.com/eternicode/bootstrap-datepicker/pull/1672#issuecomment-156554850)

PR #1674, opened 2015-11-13
Document data-api on container

fabdouglas:
Document data-api on container : input-group/input-append/input-prepend


acrobat:
Thanks @fabdouglas!

PR #1699, opened 2015-12-03
Changed translations

tigranbalayan:
May be translations was not correct.))


hebbet:
can you explain why every arrays one time more than before?

tigranbalayan:
I don't remember why I did it because I wrote this plugin about a year ago and forgot to commit.
But in more locales is the same which is this.

tigranbalayan:
I mean I remembered. For example you can use datetimepicker [0-6] format or [1-7] format for a week. That last weekday for that case. )))

acrobat:
I don't think we should accept this PR, it only adds a extra translation which duplicates the translation of "sunday" 

:-1: from me. @hebbet what do you think?

hebbet:
there are currently some translation with 8 items, but most of them only use 7, so :-1: from me too, 

but thanks for the PR.

tigranbalayan:
Thanks, but in request not only sunday's case, There are weekday and month short name changes.

acrobat:
@tigranbalayan Please create a new separate PR with the specific changes!

tigranbalayan:
OK. Thanks. I'll do it today.

Issue #1721, opened 2015-12-16
Rename "remove" function to "destroy"


tgirardi:
Bootstrap plugins use the name "destroy" for the method that removes the plugin form the web page. I suggest using the same name for standardization purposes. 

For backward compatibility, the name "remove" could be an alias of "destroy".

I can make a PR of this if you find it to be a good idea.

acrobat:
That sounds like a good idea! Please create a PR if you have the time, thanks!

tgirardi:
Nice. Work for the weekend then hahaha

acrobat:
Closing this issue. Fixed in #1732 and will be released in 1.6.0

PR #1732, opened 2015-12-23
Rename "remove" to "destroy" and create alias

tgirardi:
Rename the "remove" method to "destroy". For backward compaitbility,
create a property named "remove" that acts as an alias of "destroy".

Include unit test for the "alias behavior" (test uses sinon, so add it
to bower.json)

Issue: #1721


tgirardi:
I didn't know if I was supposed to include changes to files in `dist/` or not, so I didn't. Please correct me if I have to

acrobat:
I don't like the fact that we need a new dependency to check the alias of a method, so in my opinion we could leave that test away and revert the new dependency and just test the new method

tgirardi:
IMHO sinon is not such a bad dependency: it's only a dependency for
development and it's a very helpful library in testing environments. It
could be useful in the future.

However, if there's little chance of ever using it for other tests, then
it's a good idea to remove it.

acrobat:
Maybe you're right, I will check the features of sinon! It could be something we could use and/or extend our test suit with

vsn4ik:
:+1: for delete simon (use qunit).

tgirardi:
@vsn4ik : sinon is not an alternative to qunit nor does qunit comes with tools for stubbing, mocking and spying, that I'm aware of. The common stack for this kind of tests is qunit + sinon.

However, Sinon is completely disposable. So, for the sake of a faster acceptance of this PR, I will remove it

tgirardi:
Ready. Sinon removed.

If the version that uses sinon is preferred, then you can merge 3158e097d0eed94bb65ecb04fe8be77f60d65f8e. If not, then this PR can be merged as is.

acrobat:
Thanks @tgirardi!

PR #1735, opened 2015-12-25
Fix bug:1. function parseDate when date parts is more than format parts.

darluc:
When default endDate is set to ‘2022-12-12’ and format is set to ‘yyyy’, the endDate is parsed to current day which is not expected result.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1967,9 +1967,13 @@
 			var fparts = format.parts.slice();
 			// Remove noop parts
 			if (parts.length !== fparts.length){
-				fparts = $(fparts).filter(function(i,p){
-					return $.inArray(p, setters_order) !== -1;
-				}).toArray();
+				if (parts.length < fparts.length){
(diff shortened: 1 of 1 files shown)


vsn4ik:
@darluc Thanks for the PR, can you provide a jsfiddle or jsbin to display the problem?

Also needed test cases.

darluc:
Here is the jsbin: http://jsbin.com/bavoco/edit?html,js,output
My solution may not perfect, but the problem does exist.

I'm not familiar with javascript test cases, and I will figure out how to create one.

vsn4ik:
This is not bug.

vsn4ik:
Use `endDate: '2022'`.

acrobat:
Yes indeed startdate, enddate, etc should be added in the format that is specified in the settings. You can create a seperate PR to add some clarification to the docs about the input format of those options

Thanks!

vsn4ik:
@darluc For unification can use `endDate: new Date(2022, 11, 12)`

darluc:
@vsn4ik Thank you. Work for me.

PR #1766, opened 2016-01-21
added overview of options

EvanCarroll:
added overview of options


acrobat:
I like the idea of an overview and with the default values of each option. But maybe we should move this to a separate reference page

A bit like symfony does [here](https://symfony.com/doc/current/reference/configuration/framework.html)

EvanCarroll:
Symphony doesn't do it at all. I was looking for more of a synopsis than anything else. Copy-and-paste all of the available options into my code and comment out the ones I don't want. I always find that the best way to do it. Normally you put them all in the SYNOPSIS for the constructor.

acrobat:
I've add a quick overview table (options + default value) in e0d2377

PR #1770, opened 2016-01-26
Major bug in docs.

EvanCarroll:
This absolutely didn't work for me on the parent div. In fact, on the parent div `data-provide="datepicker"` makes it such that clicking the day does not result in the date actually being placed into the input box.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -58,8 +58,8 @@ Markup with component
 
 .. code-block:: html
 
-    <div class="input-group date" data-provide="datepicker">
-        <input type="text" class="form-control">
+    <div class="input-group date">
+        <input type="text" class="form-control" data-provide="datepicker">
         <div class="input-group-addon">
             <span class="glyphicon glyphicon-th"></span>
         </div>


acrobat:
@EvanCarroll which version of the datepicker do you use? As the example in the docs works for me in 1.5.1 (and 16.0-alpha) http://jsfiddle.net/zcvq0yhh/23/

Can you provide a jsfiddle to show case which has a problem?

EvanCarroll:
@acrobat sure, seems as if input-group's data-provide breaks if there is no `input-group-addon`. Find the example here:

http://jsfiddle.net/zcvq0yhh/29/

Also, take note at the corners. If you put data-provide=datepicker on the div the left-corners change when input field takes focus. That's a bug. Amusingly, they don't change back either when it loses focus.

EvanCarroll:
@acrobat  feedback?

acrobat:
Sorry last few weeks were quite busy. But his looks more like a bug to me then an issue with the docs. I will take a look into it, will put this issue/pr in the 1.6.1 milestone!

EvanCarroll:
Any progress on this?

EvanCarroll:
We can also see funky behavior, I believe caused by this using data-provide="daterange" on the div wrapping a date range..

http://jsfiddle.net/1Lv9cqy4/1/

Azaret:
If I'm not wrong I think that a wanted behavior, like if you initiate the picker on an input it will create a popup, but if you initiate on a div it will create the picker inside that div, that's useful if you don't want a popup, or you want to have a more complex popup like so:

image

So I think there is no bug, but an error on the documentation.

PR #1777, opened 2016-01-28
Extend beforeShowMonth event functionality to work like beforeShowDay and beforeShowYear

tpeter1985:
The event `beforeShowMonth` could only disable the cells.
I extended it with the functionality of `beforeShowDay` so you can disable, add classes and set a tooltip.
Backwards compatibility is given.

Tests are added and documentation is extended.


acrobat:
Thanks @tpeter1985!

PR #1807, opened 2016-02-26
Add options to customise datepicker popup with user defined class

YuraDubensky:
This options will be extremely useful when there is a need to have 2 date pickers with different styles on one page (like different styles for date range pickers). That way it is possible to override bootstrap styles not globally, but only for specific page/usage.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -134,6 +134,11 @@
 		else {
 			this.picker.addClass('datepicker-dropdown dropdown-menu');
 		}
+		
+		if (options.class)
+    {
+      this.picker.addClass(options.class);
+    }
 
 		if (this.o.rtl){
 			this.picker.addClass('datepicker-rtl');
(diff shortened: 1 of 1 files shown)


vsn4ik:
Also add info for docs and add test cases.

acrobat:
@YuraDubensky Can you add the docs for this option and a test?

YuraDubensky:
Yep, I will try to do so. Currently setuping git, as I don't typically work with it.

YuraDubensky:
Have added test and updated docs.

acrobat:
Can you the `class` option to the `$.fn.datepicker.defaults` array and use $.noop as default value

YuraDubensky:
Not sure if I did "rebase" in a correct way...

acrobat:
@YuraDubensky I think something went wrong during the rebase. Can you apply your changes to a new (up-to-date) branch and and open a new pr with that branch? Thanks!

PR #1824, opened 2016-03-10
originalEvent object needs preventDefault fn

bbarry:
preventDefault function expected by jQuery 2.2.1

not sure if:

    ok(evt.originalEvent.isDefaultPrevented, 'prevented original event');

should be added to test.


acrobat:
Thanks @bbarry

acrobat:
I've added the extra check in 9341da0

PR #1825, opened 2016-03-10
Meteor package

rgnevashev:


acrobat:
I'm going to close this PR for now as we already support some package managers and it looks like it is sufficient for now. We will see if we get more requests for a meteor package in the future, then we will be able to add it.

Thanks!

PR #1835, opened 2016-03-22
Generate dist files

csjoao:


acrobat:
I will try to make some time to release the next version in the next few days. Dist files will be updated on all maintained branches

PR #1855, opened 2016-04-13
Fix #1795

### LINKED ISSUE #1795: remove sourceMappingURL form .css and .min.css builds perhaps?
I think including source maps is useful for developers of bootstrap-datepicker.

However, when using your great library in large projects where your .css file is imported (because the projects don't use less e.g.), it becomes a slight nuisance to ignore/avoid all those failed requests for bootstrap-datepicker.min.css.map.

vsn4ik:


hebbet:
thanks merged.

acrobat:
@vsn4ik should we also remove the sourceMap options in the [less build step](https://github.com/eternicode/bootstrap-datepicker/blob/master/Gruntfile.js#L90)?

vsn4ik:
@acrobat for `development` is not necessary to remove. I deleted `sourceMap` only for production.

acrobat:
So this will generate the sourceMaps but don't link to them in the minified css files, if I understand correctly?

vsn4ik:
All right.

PR #1870, opened 2016-04-26
Allow user to set value that belongs to disabled range

elartix:
Feature to set picker's date that belongs to disabled range (below start date or greater than end date), when it received as a server response, but user is still disallowed to set/select such date as input


acrobat:
I don't see why this should be a standard function in the datepicker lib. As this we have the startDate option just to block users to only select dates starting from a given date. If users should be able to pick an earlier date, imo the startDate should be changed

:-1:  on this. What do you guys think? \cc @hebbet, @vsn4ik, @Azaret

hebbet:
:-1:  from here, too.

Azaret:
Tho I'll be glad too hear an use case for this feature if you mind @elartix .

PR #1878, opened 2016-05-03
Fix bug when clicking on a day that is highlighted today in the next month.

timrwood:
When the date picker is in `days` mode, and the current day is in the next month, and it is clicked, the wrong month is chosen.

image

Clicking the highlighted `3` there results in the date picked below.

image

I think this is happening because there is code written to handle clicking on the large `today` button that is also being run on the highlighted `today` day of the month. This PR prevents that code from running when clicking on the highlighted `today` day of the month, so it is treated like any other day.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1232,7 +1232,7 @@
 			}
 
 			// Clicked on today button
-			if (target.hasClass('today')){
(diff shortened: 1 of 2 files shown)


acrobat:
:+1: Looks good to merge! Will try to make some time to merge this and finish the 1.6.1 release this weekend!

acrobat:
Thanks @timrwood

Issue #1876, opened 2016-05-02
Datepicker disappears when clicking scroll bar and doesn't return when clicking in input box again.


casperson:
This only happens when you click on the scroll bar on a page. When you click, the datepicker disappears, but the input box doesn't lose focus. Now if you click back into the input box (which still has focus) the datepicker will not pop open. You have to click outside the box to blur it, then click back in. If you click outside the input box anywhere else besides the scroll bar, then the input box loses focus. It only keeps the focus on the input box when clicking right on the scroll bar.

See it in action on the demo site! [http://eternicode.github.io/bootstrap-datepicker/](url) Make sure the window is small enough to create a scroll bar though, obviously. 

1.  Click in the datepicker input box to open datepicker.
2.  Now click directly on the window scroll bar. Datepicker should disappear while input box keeps focus.
3.  Click back into the datepicker input box. 

That should do it. Just need to blur the input box, I believe. I replicated it on multiple browsers.

Azaret:
So the issue here is coming from "recent" updates of webkit. (_At least for me since we upgraded some webapp where we use datepickers from iOS5 to iOS9 recently, yeah took us some time..._)

When you click outside an input, the chain of events goes as follow:

1. `mousedown` (& `touchestart` on mobile) with where you clicked as target and bubbles.
2. `focusout` with the input as target
3. `click` with where you clicked as target and bubbles.
4. `transitionend` with always the input as target and bubbles.

The issue we're facing is that when you click on the scrollbar, the browser send `mousedown` but does not lose the focus of the input. Which is understandable since the browser want you to know that the user clicked, but not lose focus on the input, since the Mobile era we don't want to lose focus when the user is panning the page.

Since hiding the picker rely on `mousedown`, it does hide the picker because the target will be `html`. Unfortunately we can't only rely on `focusout` either because it doesn't provide the next focused target.

The idea of my PR is to catch which element have been clicked on the `mousedown` event and then wait for `focusout` to decide if we want to hide or not the picker. It's maybe not the best solution but it all I came up with for now.

We also could add `html` as an exception but I'm not really feeling like it either.

@acrobat @vsn4ik @hebbet Your thoughts are more than welcome.

stanpalatnik:
Why not just keep track if the picker is open or not, and check against it whenever we focus on the picker input?

stanpalatnik:
Created a solution that blurs the input whenever it hides the datepicker. Those actions should be coupled. What do you guys think?

[...]

PR #1885, opened 2016-05-08
Mousedown event rework - Fix for #1876

Azaret:
This should be a good fix for the issue #1876.
The idea is to store on which element the `mousedown` event and then wait for the `focusout` event to be fired, and only then test whenever the input lost focus for the picker or not.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -334,6 +334,7 @@
 				o.defaultViewDate = UTCToday();
 			}
 		},
+		_mousedown_target: null,
 		_events: [],
 		_secondaryEvents: [],
 		_applyEvents: function(evs){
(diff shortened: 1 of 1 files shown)


vsn4ik:
This is a bad fix. That's right just change mousedown to click, but it does not work with an official bootstrap-dropdown

Azaret:
I'm not sure what you're referring too.
I dont see change in behaviors except for not closing the picker when `mousedown` is fired without `focusout`.

acrobat:
@vsn4ik can you provide an example why this will not work with bootstrap-dropdown?

vsn4ik:
@acrobat in first this not fixed issue.
```
git clone -b gh-pages https://github.com/eternicode/bootstrap-datepicker.git bs-datepicker
git clone -b fix-1876-mousedown_event https://github.com/Azaret/bootstrap-datepicker.git azaret-bs-datepicker
cd azaret-bs-datepicker
npm install
grunt dist
cp -rf dist/... ../bs-datepicker/
cd ../bs-datepicker
```
update `<link href>` and `<script src>` attrubutes on `index.htm`.
open `index.html` on your browser and check fix issue.

acrobat:
@Azaret do you have time to finish this pr? See @vsn4ik comment. I don't know what extra work needs to be don so you should ping him if you need extra info!

acrobat:
@Azaret and @vsn4ik do you guys have time to pick this PR back up? Don't know exactly what still needs to be done

PR #1890, opened 2016-05-13
Capital letters in Polish translation

marioosh-net:


acrobat:
I saw that PR #1340 all names were converted to lowercase and I found this on wikipedia

> Names are generally capitalized in Polish as in English. Polish does not capitalize the months and days of the week, nor adjectives and other forms derived from proper nouns (for example, angielski "English").

[Source](https://en.wikipedia.org/wiki/Polish_orthography#Capitalization)

So I guess we should keep them lowercase

marioosh-net:
Generally You are right - in the middle of a sentence the names of months and days in the Polish language are written in lowercase, but in context of something like the calendar is absolutely correct. I am Polish ;)

acrobat:
Ok great, just wanted to check we don't override changes that should be kept! Thanks @marioosh-net

PR #1895, opened 2016-05-18
Issue #1876

stanpalatnik:
My approach is to just blur the element every time we hide the picker. Feels a little scorched earth, but it definitely fixes the issue and doesn't change any user experience(the focusout event would've happened anyway). 

I can't think of a way to write a test for this though. I think the existing tests should cover it.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -513,7 +513,7 @@
 			if (this.o.forceParse && this.inputField.val())
 				this.setValue();
 			this._trigger('hide');
-			this.inputField.blur();
+			$(this.element).blur();
 			return this;
 		},
 
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -513,6 +513,7 @@
 			if (this.o.forceParse && this.inputField.val())
 				this.setValue();
 			this._trigger('hide');
+			this.inputField.blur();
 			return this;
 		},
 


vsn4ik:
Use tabs instead whitespaces.

stanpalatnik:
That was always the case. Where do you see whitespace?

vsn4ik:
Yes, I was wrong.

acrobat:
@vsn4ik The thumbs down on the first post is that about the tabs <-> whitespace issue? or is something wrong with the suggested fix?

vsn4ik:
@acrobat Breaks standard `<input>` behavior.

stanpalatnik:
@vsn4ik This entire plugin modifies the standard `<input>` behavior.

bvauchelle:
We are using bootstrap-datepicker and stumbled upon this issue today. In our case, it seems the proposed fix is fixing the issue and we would definitely appreciate a fix (this one or another one if not approved) makes its way as soon as possible to a new release.

stanpalatnik:
@bvauchelle This proposed fix will never see the light of day since it is a little aggressive to manually blur the input on a plugin. If you find that acceptable(I do), then I recommend you merge this into your own fork.

acrobat:
@vsn4ik do you have another solution for this issue? Few people getting the same error and we should find a fix for it (if you have some time)

/cc @hebbet @Azaret any idea guys?

PR #1900, opened 2016-05-24
Add missing monthsTitle

radmax:
Czech localization was missing `monthsTitle` translation.


acrobat:
Thanks for the PR @radmax, can you just revert the identation change? So we only have the git history for the missing translations!

acrobat:
Thanks @radmax

PR #1912, opened 2016-06-07
Do not trigger change or changeDate on initialization

### LINKED ISSUE #1512: changeDate is triggered on initialization
When you initialize datepicker and the input already contains a date value, `changeDate` event is triggered. IMHO this is wrong because the date has not actually changed in that case. 

http://jsfiddle.net/hous9y5L/6/

### LINKED ISSUE #1561: Change event triggered on initialization
Hello,

EDIT: So I just found [#1512](https://github.com/eternicode/bootstrap-datepicker/issues/1512) for which this is almost a duplicate. That thread is about the ``changeDate`` event though, so I decided to post anyway. I'll let the contributors decide what to do with this issue.

I have a *suggestion* regarding the datepicker after upgrading from version 1.4 to 1.5.

I have noticed that the datepicker triggers a change event on the input at initialization.

This is a problem for us because we have change event listeners on all inputs of our form. Of course, we only want to handle user-made events. ~~We were able to work around it since all our datepickers have a special class, so that we now only listen for the plugin's ``changeDate`` event on those elements.~~ This does not really fix it, since the ``changeDate`` event is also triggered on init, but only when the original input already has a value assigned (see linked issue in EDIT on top).

I traced the event and found out that it's most likely related to [line 803](https://github.com/eternicode/bootstrap-datepicker/blob/master/dist/js/bootstrap-datepicker.js#L801) which was not there in v1.4. 

I'm not familiar with your codebase but as far as I understood, the plugins' internal ``update`` function is called when the datepicker is first initialized. 
This does not look like a bug but I think that the behavior could be optimized - it would be more intuitive and standardized to be able to listen for the common ``change`` event and I don't think it's good to trigger the ``change`` on init, since there was in fact no change.

Can anyone relate?

PS. I'll post a fiddle when I have the time, if necessary.

### LINKED ISSUE #1846: Initialising a Datepicker element calls update() which triggers a change event even if nothing has changed
The web application we implement and maintain contains a lot of forms. We have event handlers that set a global `formModified` to true whenever a form field changes. `window.onbeforeunload()` then checks for `formModified` and displays a warning when it's true.

After updating bootstrap-datepicker.js to v1.6.0 we received reports by users about these warnings appear when leaving a page even when they changed nothing.

I analysed it and saw that when initialising a Datepicker, it calls update(). And at the very end of this function, before the final `return this;` you introduced:
`this.element.change();`

This triggers our event handlers and thus sets  `formModified` to true, which then triggers the alert when a user tries to leave the page.

What is the reason for a "forced" update event in the case that the elements value has not actually changed? In my opinion this is not entirely correct.

teoulas:
Fixes #1561
Fixes #1512
Fixes #1846
... and probably more, please add them if you can.

Also includes relevant test.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -783,17 +783,21 @@
 			if (fromArgs){
 				// setting date by clicking
 				this.setValue();
+				this.element.change();
 			}
 			else if (this.dates.length){
 				// setting date by typing
-				if (String(oldDates) !== String(this.dates))
+				if (String(oldDates) !== String(this.dates) && fromArgs) {
 					this._trigger('changeDate');
+					this.element.change();
+				}
 			}
-			if (!this.dates.length && oldDates.length)
+			if (!this.dates.length && oldDates.length) {
 				this._trigger('clearDate');
+				this.element.change();
+			}
 
 			this.fill();
-			this.element.trigger('change');
 			return this;
 		},
 
--- a/tests/suites/events.js
+++ b/tests/suites/events.js
@@ -1,3 +1,29 @@
+module('Events on initialization', {
+    setup: function(){
+        this.input = $('<input type="text" value="31-03-2011">')
+            .appendTo('#qunit-fixture')
+    }
+});
+
+test('When initializing the datepicker, it should trigger no change or changeDate events', function(){
+    var triggered_change = 0,
+        triggered_changeDate = 0;
+
+    this.input.on({
+        change: function(){
+            triggered_change++;
+        },
+        changeDate: function(){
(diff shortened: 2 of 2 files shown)


acrobat:
@hebbet, @vsn4ik @Azaret feedback? Looks good to me and does indeed fixing a good amount of tickets

vsn4ik:
LGTM!

teoulas:
Hey guys,

I've just rebased and fixed the conflicts. I would appreciate if we merged this soon.

acrobat:
Thanks @teoulas

teoulas:
Awesome, thanks! 👍

Issue #1921, opened 2016-06-17
Change the dependency on jQuery not to allow jQuery 3?


linsolas:
Hello,

Bootstrap 3 is [not compatible with jQuery 3](https://github.com/twbs/bootstrap/issues/16834), and may not be corrected to support that version.
Unfortunately, bootstrap-datepicker allows [any version of jQuery superior to 1.7.1](https://github.com/eternicode/bootstrap-datepicker/blob/master/package.json#L27-L29):

```
  "dependencies": {
    "jquery" : ">=1.7.1"
  }
```

Shouldn't bootstrap-datepicker restrict the jQuery version, something like that:

```
  "dependencies": {
    "jquery" : ">=1.7.1 <3.0.0"
  }
```

Thanks

acrobat:
Yes that's indeed a too open constraint! Can you open a PR for that? If you open a PR please also update the constraint in `bower.json` and `composer.json`

linsolas:
Thanks. I wanted to have a collaborator opinion before. I've made the #1922 for that.

acrobat:
:+1:. Fixed in #1922

PR #1922, opened 2016-06-17
Do no allow jquery 3

linsolas:
PR to solve the issue of #1921 since Bootstrap 3 is not compatible with jQuery 3 (https://github.com/twbs/bootstrap/issues/16834).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -6,7 +6,7 @@
   ],
   "license": "Apache-2.0",
   "dependencies": {
-    "jquery" : ">=1.7.1"
+    "jquery" : ">=1.7.1 <3.0.0"
   },
   "ignore": []
 }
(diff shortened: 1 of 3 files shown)


acrobat:
Thanks for the quick fix @linsolas!

glen-84:
This should be reverted, as Bootstrap 3 *is* now compatible with jQuery 3.

niemyjski:
What's the status of jquery3 support?

acrobat:
From 1.6.2+ we have restored jauery 3 support. See the changelogs for linked issue about jquery 3!

niemyjski:
I'm using the latest version and I think the bower def hasn't been updated
as I'm prompted for a resolution.

Thanks
-Blake Niemyjski

PR #1926, opened 2016-06-21
add .map-files

mrohnstock:
webbrowsers do possibly look for these files, so let's add them, too.


mrohnstock:
Or should these map-files get added to the `styles`-block instead of `files`?

Issue #1274, opened 2015-02-23
daylight saving time


jpsduarte:
There is a problem while rendering the dates when daylight saving time starts and ends. For example, here in Brazil, daylight saving time started in Octuber 19, 2014 and in this day we were unable to select  the date we wanted. If we try to do a console log in Chrome with "new Date('Octuber 19, 2014 00:00:00')", this give us Octuber 19, 2014 01:00:00. It adds 1 hour to hour date because, theoretically, this hour doesn't exist. Can you help me?

acrobat:
there is a related issue about this problem (or something like this) here #844. But can you provide a jsfiddle example showing the bug (you can use [this base jsfiddle](http://jsfiddle.net/g02L8z8y/) to start with)? That way we can get as much info as possible to solve this issue

FabioReis:
I had exactly the same error. I´m also from Brazil, and Saturday is the last day of our daylight time(02/21/2016), so in this case i´m on GMT -03:00 and because of the daylight time our timezone is currently GMT -02:00. I did as you asked and i created a fiddle that shows exactly the code i have. Of course in order for this example to work you need to change your timezone. Thanks!
http://jsfiddle.net/tzgfbobs/

bertnz:
Looks like the exact same issue as here
https://github.com/dalelotts/angular-bootstrap-datetimepicker/issues/219

PR #1928, opened 2016-06-22
UTC to local date conversion incorrect on daylight saving boundary

growlycode:
Candidate fix for 
https://github.com/eternicode/bootstrap-datepicker/issues/1274

This PR contains the UTC -> local date conversion fix applied to dalelotts/angular-bootstrap-datetimepicker to remedy the same issue in their repo.
https://github.com/dalelotts/angular-bootstrap-datetimepicker/commit/74488250e4e08070b4bb67fdd541009f3ad113fc

Can anyone who knows this codebase comment on whether this would also be an issue in the reverse direction (_local_to_utc)? That method also uses getTimezoneOffset().

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -529,7 +529,7 @@
 		},
 
 		_utc_to_local: function(utc){
-			return utc && new Date(utc.getTime() + (utc.getTimezoneOffset()*60000));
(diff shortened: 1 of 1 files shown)


acrobat:
@Azaret I think you had some experience with the whole timezone/datetime problem?

/cc @vsn4ik

vsn4ik:
@growlycode in this case `_zero_time` function is almost the same. 
Maybe try and leave only `_utc_to_local` and `_local_to_utc`.

allentong:
I tested this in London TZ (UTC+00:00) and the issue isn't resolved properly.

PR #1946, opened 2016-07-14
Fix data-date-dates-disabled attribute to accept a comma-separated list

maxfurman:
Setting multiple datesDisabled through a data attribute is currently broken, because the _process_options function assumes that if that value is not already an array it is only one datestring. This is not always the case; if you try to pass multiple dates to the data-date-dates-disabled attribute they will be interpreted as one long string and fail to parse as dates. Fixed by calling `.split(',')` on the raw value. Strings without commas will simply be wrapped in an array, which is the current behavior.


acrobat:
Thanks @maxfurman

PR #1949, opened 2016-07-20
Fixing the width of the calendar

andyj:
When RTL is TRUE the width of the window is stretched


acrobat:
@andyj can you provide a before and after jsfiddle because if I try the patch the width is not enough to cover the correct width of the picker

PR #1953, opened 2016-07-22
added "changeViewMode" event; also adds the current `viewMode` to events

AubreyHewes:
Adds a "changeViewMode" event (triggered by navArrowsClick). Also adds the current `viewMode` to any triggered events.

So now you can watch `changeViewMode` and depending on the `event.viewMode` and the `event.date` you can do something..

The reasoning was that on a `viewMode` change I wanted to be able to get dates by viewMode range from a backend and then redraw the component ...


acrobat:
Thanks @AubreyHewes

PR #1954, opened 2016-07-23
added `today` class to all view mode levels (todayHighlight)

AubreyHewes:
If `todayHighlight` is set. This adds the `today` class to all view modes.. 
Also changes the _today button_ class to another, though keeping the old for BC. This is because it conflicted with _today button_ click logic.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -882,7 +882,10 @@
 		},
 
 		_fill_yearsView: function(selector, cssClass, factor, step, currentYear, startYear, endYear, callback){
-			var html, view, year, steps, startStep, endStep, thisYear, i, classes, tooltip, before;
+			var html, view, year, steps, startStep, endStep, thisYear, i, classes, tooltip, before, today;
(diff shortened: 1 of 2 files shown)


acrobat:
@AubreyHewes maybe we should add another class for the highlight in all viewmodes and keep the today class on the button to avoid any bc breaks. And make a note for the 2.0 version to use the `btn-today` class instead of `today` on the button

AubreyHewes:
@acrobat soz was on a little holiday... i.e. something like `btn-now` on all views and `today` on day view _today_ (keeping bc) ?

Azaret:
I'm not in favor of multiplying class, there is no real need for this right now.
Change the click logic for 
```javascript
if (target.hasClass('today') && !/[^o]day|month|year|decade|century/.test(target.prop("class"))) {}
```

I think we should keep classes for styling, with few exceptions.
The today button is easily distinguishable by being the only element in the footer.
In the future we might simply store the element in a variable, it will help to keep performances down.
Not that the plugin has bad performances, but we can do better.

PR #1958, opened 2016-07-28
Add support jQuery 3. Bootstrap 2 still available (composer.json)

vsn4ik:
Fixed #1948.

PR #1964, opened 2016-08-03
Updated html template in JS to work with jQuery 3

erestor:
Explicit "display:block" style is required on datepicker-days, datepicker-months etc. elements in order for jquery .show() to add style="display:block" on showMode() which overrides the default style
.datepicker > div {
    display: none;
}


acrobat:
Thanks @erestor for the fix! I've merged #1958 which fixed the jquery3 issue it bit cleaner! Jquery 3 support is restored in master (1.7-dev) and 1.6.2

PR #1968, opened 2016-08-04
Remove listen `changeDate` after destroy DateRangePicker

lusever:
Need for correct recreate DateRangePicker on same inputs.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1567,6 +1567,7 @@
 		},
 		destroy: function(){
 			$.map(this.pickers, function(p){ p.destroy(); });
+			$(this.inputs).off('changeDate', this.dateUpdated);
 			delete this.element.data().datepicker;
 		},
 		remove: alias('destroy')


js/bootstrap-datepicker.js
```diff
@@ -1567,6 +1567,7 @@
 		},
 		destroy: function(){
 			$.map(this.pickers, function(p){ p.destroy(); });
+			$(this.inputs).off('changeDate', this.dateUpdated);
```
vsn4ik: `this.dateUpdated` !== `$.proxy(this.dateUpdated, this);`

lusever: Yes. But it is work for jQuery. From [docs](http://api.jquery.com/off/):

> When jQuery attaches an event handler, it assigns a unique id to the handler function. Handlers proxied by jQuery.proxy() or a similar mechanism will all have the same unique id (the proxy function), so passing proxied handlers to .off may remove more handlers than intended. In those situations it is better to attach and remove event handlers using namespaces.

lusever: И чего?

vsn4ik: https://github.com/eternicode/bootstrap-datepicker/blob/master/js/bootstrap-datepicker.js#L2001 используется нэймспейс. Я считаю что лучше использовать нэймспейс, даже jQuery советует.

lusever: I changed the way of removing events to namespace.

vsn4ik:
You need to add namespace here (https://github.com/eternicode/bootstrap-datepicker/blob/master/js/bootstrap-datepicker.js#L1513) and add testcase.

lusever:
I added namespace.
@vsn4ik prompt me please place where to create the testcase correctly.

acrobat:
@vsn4ik I think we can't add a namespace now, this would mean a BC break. Why is the namespace necessary ?

vsn4ik:
@acrobat Let's roll out without tests. Namespace is not necessary.

acrobat:
@lusever Can you remove the `.datepicker` namespace on both events? If that still fixes your issue the pr is good to go!

lusever:
@acrobat I removed namespaces.

lusever:
If I use the code which is currently:
```javascript
$('.inputs input').on('changeDate', myFunc)
$('.inputs').datepicker({inputs: $('input')})
$('.inputs').data('datepicker').remove(); // removed myFunc too
```
I think the original version is more correct. It is like in `_unapplyEvents`. 
I think it is necessary to restore the original version. Then `myFunc` will not be removed.

acrobat:
So the changes in this PR are not correct @lusever?

lusever:
@acrobat now is correct.

acrobat:
Thanks @lusever!

PR #1973, opened 2016-08-09
Fix RTL layout

nwhittaker:
I believe this fixes the issue where RTL layouts blow out the width of the picker. Specifically, the left side of the picker appears to extend to the left side of the viewport.

* Fixes RTL width.
* Adjusts position of arrow/caret to the right side of the picker.
* ~~Adds build artifacts.~~


nwhittaker:
Not sure if it was helpful to include the build artifacts. Here's the core change: https://github.com/nwhittaker/bootstrap-datepicker/pull/1/commits/ef6244b6c8f14dbe18b0fe26c74ccda3dd11d457.

acrobat:
Thanks for the pr @nwhittaker! But can revert the changes on the dist files and just keep the core changes to the js and less files

Thanks!

nwhittaker:
>But can revert the changes on the dist files and just keep the core changes to the js and less files

Sorry about that -- wasn't sure what the procedure was. Should look a little better now.

acrobat:
Thanks @nwhittaker, can you provide a jsfiddle example to show that the issue is resolved? As I don't seem to see a difference with the current code

nwhittaker:
Looks like I had a problem with the specificity of the CSS selector. I'll push a fix.

In case it's still useful, here's a [fiddle of the existing behavior](https://jsfiddle.net/n4u6sjor/7/), and [one showing the issue resolved](https://jsfiddle.net/urmoup7r/7/).

nwhittaker:
How are we looking?

nwhittaker:
@acrobat, what are your intentions with this PR?

timmorey:
This fixes an issue that I've been running into - @acrobat can we get a merge here?

musabsudan:
@acrobat could you please merge this?

acrobat:
Sorry for the late answer, I've been on vacation the last few days. Pr is merged in master and will be included in 1.7.0

Thanks @nwhittaker

Me1o:
this.picker.css({
		top: o,
                left: n,
		right: n,
                zIndex: j
            });

this worked for me :)

Issue #1978, opened 2016-08-16
Add methods `getViewDate`, `setViewDate` to public api


jelhan:
Would you accept a PR which adds `getViewDate` and `setViewDate` methods to public api? Implementation should follow `defaultViewDate` even so at least I didn't expected such a syntax.

So in detail I like to add this public api methods:

`$().datepicker('getViewDate');` returns current viewDate as a plain object having keys `year`, `month`, `day`. 

`$().datepicker('setViewDate', { year: 2000, month: 0, day: 1 });` updates datepicker with view date ''1 January 2000''. If keys are missing they should default to the same as `defaultViewDate`. So `$().datepicker('setViewDate', {});` would update datepicker with today as view date.

It's already possible to manipulate viewDate by using private api method `_setDate` with a `Date` object as first and `'view'` as second parameter. So implementation is easily possible.

This new public api methods would allow implementing linked datepickers without using private api. Linked datepickers could be used to implement "showing multiple month" feature which is heavily requested (#1154 and many duplicates). I've written a proof of concept to show it's actually working: https://jsfiddle.net/v6q2rxs0/

`getViewDate` isn't really necessary therefore but should be added for consistency.

Would love to get some feedback before starting implementation.

TerjeBr:
Any updates on this?
This would be a really useful feature.

PR #1982, opened 2016-08-18
adds `updateViewDate` option

jelhan:
Implementation details are present in updated docs and added test. Default value does not change any existing behaviour.

I think this option will be helpful if you interact with bootstrap-datepicker via code. It allows you a more granular control how bootstrap-datepicker reacts on changes. It will be really powerful together with `setViewDate` method suggested in #1978.

Please have a review on syntax used in docs. I'm not quite sure if lists are supported and wasn't able to build docs to verify.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -770,14 +770,16 @@
 			}, this), true);
 			this.dates.replace(dates);
 
-			if (this.dates.length)
-				this.viewDate = new Date(this.dates.get(-1));
-			else if (this.viewDate < this.o.startDate)
-				this.viewDate = new Date(this.o.startDate);
-			else if (this.viewDate > this.o.endDate)
-				this.viewDate = new Date(this.o.endDate);
-			else
-				this.viewDate = this.o.defaultViewDate;
+			if (this.o.updateViewDate) {
+				if (this.dates.length)
+					this.viewDate = new Date(this.dates.get(-1));
+				else if (this.viewDate < this.o.startDate)
+					this.viewDate = new Date(this.o.startDate);
+				else if (this.viewDate > this.o.endDate)
+					this.viewDate = new Date(this.o.endDate);
+				else
+					this.viewDate = this.o.defaultViewDate;
+			}
 
 			if (fromArgs){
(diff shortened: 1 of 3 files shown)


acrobat:
@jelhan can you provide a jsfiddle example of a possible usecase so I can test the pr? Thanks!

jelhan:
@acrobat Here you go: https://jsfiddle.net/pkpx6baw/ Also note that tests are included in PR.

acrobat:
Thanks @jelhan

js/bootstrap-datepicker.js
```diff
@@ -1273,7 +1279,7 @@
 		_setDate: function(date, which){
 			if (!which || which === 'date')
 				this._toggle_multidate(date && new Date(date));
-			if (!which || which === 'view')
+			if ((!which && this.o.updateViewDate) || which === 'view')
```
vsn4ik: Hmm... Maybe `(!which || which === 'view') && this.o.updateViewDate)`?

jelhan: @vsn4ik `(!which || which === 'view') && this.o.updateViewDate` would make it impossible to change `viewDate` if `updateViewDate` is `false`. Perhaps `_setDate` should be better documented.

Azaret:
This feature actually breaks the default and expected behavior of the picker.
It should be set to `false` by default.

Moreover can we have some use cases on why anyone would want to set this to `true` ?
I'm not convinced.

jelhan:
@Azaret Could you please provide a case where it breaks something? It should not change any existing behaviour if `updateViewDate` is set to `true` which is default.

Azaret:
Change the month, call `update` and the picker goes back to the current month.
This issue is that when set to `true` `viewDate` is not updated anymore so any call made to `update`, `setDates`, ... will reset the view. I'm not sure in which case we would want sure behavior.

Now maybe there is a side effect with another change, but I noticed that turning that `false` solve the issue of the view being reset.

jelhan:
@Azaret I'm not quite sure what you mean. I see same behavior in 1.6.4: https://jsfiddle.net/1gv2vwau/

Issue #1980, opened 2016-08-17
Switching month when setting disabled dates dynamically


fossecode:
### Expected behaviour
I am making a calendar which loads disabled dates when changing months. Every time the month switch, an ajax request is sent for the previous month, the current month and the next month. Nothing should change on the calendar except of the disabled days.

### Actual behaviour
The disabled days are disabled correctly, but the month visible in the calendar jumps back to the `defaultViewDate.`

### Example code
`$('.appointment-datepicker').datepicker('setDatesDisabled', disabledDates);`

jelhan:
Could you please provide some more information or write failing tests? I have tried to reproduce in #1983 and some manual testing but wasn't able to do so.

Worthy7:
I have just come accross a similar bug. I don't fully understand how this works, but the error is because of 
if (this.dates.length)
				this.viewDate = new Date(this.dates.get(-1));

on line 794. The original this.viewDate is getting overwritten by the previous date, because the "this.dates" array has not been updated yet.

So to be clear, the problem I THINK is that the new date is not in the this.dates array, which is why on "update" the date reverts. So is this a bug with the way events have been written?

Worthy7:
Ok I worked it out.

What's happening is, the functions around like 650~700, which manges all the methods, are using this.update(), which actually clears the current selection and resets it to default (like, today).
This should be this.fill() instead. Ill let the author look into this as I've just spend 2 days of work on this and have to move on. Hope this helps someone.

jelhan:
@Worthy7 Thanks for sharing more details about the bug. I was able to write a failing test.

Worthy7:
Great! Now fix it :)

Azaret:
This is actually related to our conversation on #2046 and #1982.
We need some serious investigation on this jump back issue.

dhakan:
I encountered the same issue needing to disable dates based on an AJAX request. I worked around the issue of the date reverting back by using a call to the internal private function `_setDate` right after the call to `setDatesDisabled`. 

First: 
`$('.datepicker-selector').datepicker('setDatesDisabled', datesToBeDisabled);`

and then, secondly:
`$('.datepicker-selector').datepicker('_setDate', dateParam, 'view');`

where `dateParam` is a date within the currently viewed month.

Not entirely sure if this does not cause any side effects.

jelhan:
[...]

PR #1983, opened 2016-08-18
add tests for setDatesDisabled function

jelhan:
Tried to reproduce #1980 in tests but wasn't able. But tests ware always good. :-)


acrobat:
Thanks @jelhan

PR #1984, opened 2016-08-18
Doc: format for endDate and startDate

### LINKED ISSUE #1977: Format for setEndDate and setStartDate?
Documentation doesn't mention which format is expected for `setEndDate` or `setStartDate`.
It refeers to endDate option, which doesn't either explain the required format.

I do not want to base endDate and newDate on the current date, but on a specific date. For example:

```
Fri Aug 19 2016 00:00:00 GMT+0100 (GMT Daylight Time)
```

jelhan:
Fixes #1977

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -197,9 +197,11 @@ If false the datepicker will not show on a readonly datepicker field.
 endDate
 -------
 
-Date.  Default: End of time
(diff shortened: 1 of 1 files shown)


acrobat:
Thanks @jelhan! I've backported the fix to 1.6 so the current stable docs are also up to date

PR #1994, opened 2016-08-29

chenyitian:
fix bug when setting arrow template with html, event target error.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1242,7 +1242,7 @@
 
 		// Clicked on prev or next
 		navArrowsClick: function(e){
-			var target = $(e.target);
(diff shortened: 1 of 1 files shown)


acrobat:
Can you provide a jsfiddle with the current behavior (which is broken?)?

Azaret:
I'm not sure but I guess we might have a issue since we allows custom html for arrows.
I guess that if you inject some tags in the arrow, when you will click on the arrow `e.target` will equals the injected tag and not the `.prev` `.next` containers.

acrobat:
I'm closing this PR as @vsn4ik fixed this bug and added a test for it. The fix is available in master and in the next 1.7.0 RC/stable release. See #2116

PR #2002, opened 2016-09-04
[Doc] Keyboard navigation is not supported for embedded / inline type

### LINKED ISSUE #1263: Keyboard navigation broken on embedded / inline
On the demo page of the Bootstrap Datepicker, when selecting the embedded / inline mode, the keyboard navigation with the arrows does not work. It does not work either on the website I'm working on. I imagine someone would have reported the issue already but I could not find it in the this issue section.

jelhan:
[Doc] Keyboard navigation is not supported for embedded / inline type at all.

Fixes #1263

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -320,6 +320,8 @@ Boolean.  Default: true
 
 Whether or not to allow date navigation by arrow keys.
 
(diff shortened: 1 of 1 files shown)


acrobat:
Thanks @jelhan

Issue #1787, opened 2016-02-12
z-index not working on IE11


michael-x:
Using eternicode bootstrap-datepicker version 1.6.0 / jQuery 2.2.0 on IE 11.

I've got a page with a navbar which is fixed at the top (navbar navbar-fixed-top).

When the datepicker opens up near the top it is displayed underneath (in z-direction) the navbar. This is kind of expected as the navbar has got z-index 1030, the default zIndexOffset of the datepicker is 10 and I don't have any z-index set on any of the parents. But, when I set the zIndexOffset to 1500 everything works on Chrome, but nothing happens on IE11.

What I've noticed is that in Chrome the appropriate z-index is added to the style attribute of the element. IE11 does no such thing. It only applies the z-index of the .dropdown-menu which is 1000, no matter whether zIndexOffset is set or not.

Not sure, if this is a datepicker, jQuery or IE issue.

hebbet:
can you test if it works with 1.5.1?

michael-x:
I've compared the two script files and noticed that the place function, responsible for calculating the z-index, hasn't changed between 1.5.1 and 1.6.0.

I've also had a look at the [demo page](http://eternicode.github.io/bootstrap-datepicker/) which seems to be using version 1.5.0. It is also using a navbar which is fixed to the top. When you resize your IE to reduce the height to about the height of the datepicker, then you can also see the same behavior.

allenpiscitello:
The source of the problem is here:

	var parentsZindex = [];
			this.element.parents().each(function(){
				var itemZIndex = $(this).css('z-index');
				if (itemZIndex !== 'auto' && itemZIndex !== 0)
				    parentsZindex.push(parseInt(itemZIndex));
			});
		    var zIndex = Math.max.apply(Math, parentsZindex) + this.o.zIndexOffset;

itemZIndex is a string of "0" in Chrome and other browsers, which makes it accidentally work.  IE interprets it as a number.  This means it will skip any element of an offset of 0.

The parentsZIndex comes back as an empty array for IE, which then has a max value of -Infinity.  Thus the offset is not applied.  The way to fix this is to initialize parentsZIndex with [0] and to add parseInt(itemZIndex) when comparing to 0.

allenpiscitello:
I have submitted the following pull request.

https://github.com/eternicode/bootstrap-datepicker/pull/2008

vsn4ik:
Fixed in next release! Thanks @michael-x!

PR #2008, opened 2016-09-15
Fixed z-index calculation error on IE11

allenpiscitello:
In IE11, the zIndex is returned as a number, which has parentsZIndex returned as an empty array.  This empty array has a max value applied, which is -Infinity, which causes the zIndexOffset to not be applied.

Chrome and other browsers accidentally get correct behavior since the zIndex property is returned a string, which makes !== 0 return true.

To unify the behavior, I added parseInt to the comparison, and set a default value of the parentsZIndex to 0, so that the max value is always at least 0.


allenpiscitello:
I know the guidelines say to have a test.  This is a reported issue here: https://github.com/eternicode/bootstrap-datepicker/issues/1787

I'm not a big JS guy, but hopefully someone can run with this.  It's a fairly straightforward fix.

vsn4ik:
Thanks @allenpiscitello!

acrobat:
@vsn4ik is this covered by tests? Can you maybe provide a test if you have the time?

PR #2012, opened 2016-09-19
format yyyy-mm-dd shall return 0001-01-01 instead of 1-01-01

michael-dev:
Currently, 4-digit year with format yyyy is not enforced. Same goes for yy being 2-digit.

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1913,9 +1913,17 @@
 				m: date.getUTCMonth() + 1,
 				M: dates[language].monthsShort[date.getUTCMonth()],
 				MM: dates[language].months[date.getUTCMonth()],
-				yy: date.getUTCFullYear().toString().substring(2),
+				yy: date.getUTCFullYear() % 100,
 				yyyy: date.getUTCFullYear()
 			};
+			if (val.yyyy < 10) {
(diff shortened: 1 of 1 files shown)


js/bootstrap-datepicker.js
```diff
@@ -1913,9 +1913,17 @@
...
 				yyyy: date.getUTCFullYear()
 			};
+			if (val.yyyy < 10) {
+				val.yyyy = (val.yyyy < 10 ? '000' : '') + val.yyyy;
+			} else if (val.yyyy < 100) {
+				val.yyyy = (val.yyyy < 100 ? '00' : '') + val.yyyy;
+			} else if (val.yyyy < 1000) {
+				val.yyyy = (val.yyyy < 1000 ? '0' : '') + val.yyyy;
+			}
+			val.yy = (val.yy < 10 ? '0' : '') + val.yy;
```
Azaret: Would be easier like this:
```javascript
val.yyyy = "0000".substring(0, 4 - val.yyyy.toString().length) + val.yyyy
```

deltab: or `("0000" + val.yyyy).slice(-4)`.

michael-dev: Maybe just fix it yourself?

Issue #2015, opened 2016-09-21
Calendar related-title issue in Chinese env when bootstrap-datepicker.zh-CN.min.js was included in the page


pei-han:
Look at the images below,easily know what i say this issue.
### Expected behaviour
![image](https://cloud.githubusercontent.com/assets/1300292/18706892/59e0d9f8-8026-11e6-92e2-c1c3a7016585.png)

### Actual behaviour
![image](https://cloud.githubusercontent.com/assets/1300292/18706869/43b96a5a-8026-11e6-846e-e37196f97c65.png)

### Datepicker version used

v1.7.0-dev

![image](https://cloud.githubusercontent.com/assets/1300292/18706969/9f73be04-8026-11e6-8302-bdbc9e0bf0a3.png)

Azaret:
Hi,

Actually I can't reproduce this issue in latest.

PR #2016, opened 2016-09-21
delele invalid format

pei-han:
Cause format property only allow d, dd, D, DD, m, mm, M, MM, yy, yyyy and seperator /
and charactars else are invalid.
 #2015


Azaret:
I'm closing this PR since it is not a solution to the related issue.

pei-han:
why not?

PR #2028, opened 2016-09-30
[l10n]Add breton translation

tornoz:
Add translation file for Breton language (Celtic language spoken in Brittany)


acrobat:
Thanks @tornoz

PR #2029, opened 2016-10-02
"rangeTwoInputs" mode added

throwable-one:
Hi. This feature I did for my own project, but I want to share it with you.
You can read doc, but here is how it works: 
* User inserts 2 inputs and provides ``rangeTwoInputs`` option
* Client may click any input and then choose first date. First date always goes to first input. Then, second date should be chosen which goes to second input. So, there is a range which is displayed visually. 

Please note this is *not* how it works in your version. Try to click on first input, choose first date and then choose second. Instead of range, second date is copied to both fields. I am note sure if it is bug or feature, but ``rangeTwoInputs`` looks cleaner for me.

Thank you for your datepicker)

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -385,6 +385,13 @@ A space-separated string consisting of one or two of "left" or "right", "top" or
 "auto" triggers "smart orientation" of the picker.  Horizontal orientation will default to "left" and left offset will be tweaked to keep the picker inside the browser viewport; vertical orientation will simply choose "top" or "bottom", whichever will show more of the picker in the viewport.
 
 
+rangeTwoInputs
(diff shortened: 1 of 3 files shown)


Azaret:
If I'm correct, that's basically what the DateRange component does.
I understand that at the moment DateRange is not really nice to use (it's planned to improve it), but imo this kind of feature should be in the DateRange class, not in the single picker one.
That's a 👎 for me.
@acrobat @hebbet @vsn4ik What do you think guys ?

acrobat:
I'm also :-1:  for now. We first have to cleanup and improve the current daterange before we should consider new features around this component

PR #2031, opened 2016-10-02
Removed reference to stefan petre's

acrobat:
Our codebase has diverged a lot from the original code. So further references aren't necessary. Discussed with @eternicode 

From 1.7.0+ there won't be any references anymore


Azaret:
RIP 🍕 
I'm not much sure about this.
First, while our code did evolved in 3 years, jQuery call, methods, events and options and event the plugin name are still the same.
Then, it offers visibility on Google, since eyecon.ro is still up. If someone if looking up for eyecon or stefan petre, maybe because they didn't updated the plugin since few years, our fork is referenced next to stefan's in Google.

At the end I don't care much to be honest, just that I guess keeping it have a bit of benefice and no harm I guess.

Azaret:
I would let some notice in the readme and the JS header.
Tho I agree on cleaning the rest ;)

Azaret:
Also this PR contains non related commit, this is evel @acrobat 😈 😄

acrobat:
@Azaret If you want you can open a new PR with some additions to the readme and jq header! Would not harm for visibility indeed!

Issue #1856, opened 2016-04-14
Serbian locales -> "sr" code does not belong to Serbia


usarskyy:
Hello,

repository contains the following Serbian locale files:
- bootstrap-datepicker.rs.js
- bootstrap-datepicker.rs-latin.js
- bootstrap-datepicker.sr.js
- bootstrap-datepicker.sr-latin.js

Accordingly to [Wikipedia](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) "SR" code does not belong to Serbia, it is [Suriname](https://en.wikipedia.org/wiki/Suriname)

Fix of this issue is quite simple: either to update bootstrap-datepicker.sr.\* files with correct translation, or drop them at all.

acrobat:
The list you are looking at is the ISO list of country codes (ISO 3166) we use [ISO 639](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) (the language code iso standard) to identify the language codes. But "rs" should be deprecated (and removed in 2.0) as it's an incorrect language code!

acrobat:
I've added the deprecation notice in 0ae9678. Translations will be removed in 2.0

vsn4ik:
Thanks @usarskyy!

PR #2036, opened 2016-10-09
Added a function to indicate deprecations

acrobat:
This way we can add deprecation notices. Users of 1.x will know what features/methods/locales are deprecated and removed in 2.0. This will make the upgrade a bit easier as they can change their code usages beforehand.

The result:
deprecation


Azaret:
👍 Don't forget #1856

js/bootstrap-datepicker.js
```diff
@@ -2034,3 +2038,13 @@
 	});
 
 }));
+
+function deprecate(msg) {
+	var console = window.console;
+	if (console && console.warn) {
+		console.warn('DEPRECATED: ' + msg);
```
vsn4ik: Spaces with tabs!

js/bootstrap-datepicker.js
```diff
@@ -2034,3 +2038,13 @@
 	});
 
 }));
+
+function deprecate(msg) {
```
vsn4ik: Global scope is bad

acrobat: How should I declare this function?

js/bootstrap-datepicker.js
acrobat: I leave it for now as this whole file is a mixup of spaces vs tabs. But we will clean this up and move to spaces when we move to 2.0

js/bootstrap-datepicker.js
Azaret: Or we just copy paste it for now in locale...

Azaret: Ooorrr: `$.fn.datepicker.deprecate = function (){};`

js/bootstrap-datepicker.js
vsn4ik: I am sorry, my bad.

Azaret: Not your fault, be sorry for the poor souls that committed mixed content... and the one that merged it 👿

js/bootstrap-datepicker.js
acrobat: maybe it's the best solution to move it to `$.fn.datepicker.deprecate`?

vsn4ik: I like `$.fn.datepicker.deprecate`.

acrobat: ok :+1: I will update the pr!

js/bootstrap-datepicker.js
acrobat: The one that merged it would probably be me :p

js/bootstrap-datepicker.js
acrobat: I've tried to move the function to `$.fn.datepicker.deprecate` but then I can't use the function in the alias method. So all tests fail, so where should I declare the method?

Azaret: How? I don't see why you could not ?

acrobat: I've pushed the change so you can see the test result on travis and how I declared the function!

Azaret: The issue is that `alias` is executed when the datepicker is injected into jQuery.
IMO it should be executed when any of the aliases is called, not before.
I can PR you a fix for alias then you can rebase on it.

Azaret: Or you should be able to do it this way:
```javascript
	function alias(method, deprecationMsg){
		return function(){
			if (typeof deprecationMsg !== 'undefined') {
				$.fn.datepicker.deprecated(deprecationMsg);
			}
			return this[method].apply(this, arguments);
		};
	}
```

acrobat: yes if you want op a pr against `acrobat:deprecations`!

Azaret: I don't event...
Gonna check latter how to do such thing ^^

acrobat: @Azaret do you maybe have some time soon to push an update for the `deprecated` method? Or just post here a small snippet on how we should do this and I will add it!

vsn4ik:
Thanks @acrobat!

PR #2042, opened 2016-10-24
Allow overriding the surrounding HTML template per picker instead of globally

programcsharp:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| License         | MIT

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -414,6 +414,12 @@ Number, String.  Default: 0, "days"
 
 The view that the datepicker should show when it is opened.  Accepts: 0 or "days" or "month", 1 or "months" or "year", 2 or "years" or "decade", 3 or "decades" or "century", and 4 or "centuries" or "millenium".  Useful for date-of-birth datepickers.
 
+template
+---------
+
+Object. Default: null
+
+Allow overriding the datepicker surrounding HTML template per picker, vs. only once in DPGlobal.template.
 
 templates
 ---------
@@ -545,6 +551,7 @@ orientation                  'auto'
 showOnFocus                  true
 startDate                    -Infinity
 startView                    0 'days' (current month)
+template
 templates
 title                        ''
 todayBtn                     false
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -113,7 +113,7 @@
 			this.component = false;
 		this.isInline = !this.component && this.element.is('div');
 
-		this.picker = $(DPGlobal.template);
+		this.picker = $(this.o.template || DPGlobal.template);
 
 		// Checking templates and inserting
 		if (this._check_template(this.o.templates.leftArrow)) {


Azaret:
Hi, Do you have some use cases?
I'm sure I see the value of such feature.

programcsharp:
In my case, I want to add some extra chrome around the datepicker, but only for some of them, not all of them.

For example, I'm working on an availability picker right now that displays holidays and time that has already been selected. I've added some custom classes using onBeforeShowDay to shade days on the calendar, and mark half days with half slashes, but I want to display a legend to the user to show what's going on.

Overriding the template allows me to do this:

image

Azaret:
Well, that's a good exemple, now there is my issue with this:

First the usability is not great, since you've to include back mostly at least all the default template in your option in order to do what you want and not break the picker. As a dev I won't be happy with a bunch of html in the middle of my javascript. It looks too heavy handed for just be able to inject some html.

Then it is not sustainable, if in the next release we have some modification in the template, we will give a bad time to anyone that is overriding the default template and want to update the library.

Now I think we can come up with a more appropriate solution. One that comes in mind right away would be to add an event when the picker is generated and to send in the event arguments the DOMElement of the picker.
Then you will be able to do something like:

```javascript
$('#datepicker').on('generated', function (element) {
  $(element).append(...);
}
```

programcsharp:
Right, yeah, I'm open to finding a better solution that's more generic. This is a one-liner fix that gave me the power I needed.

The event could maybe work, but what about construction? I mean, the event would be fired during the initial datepicker call, before you attached.

In other words, if you do:

```$('#datepicker').datepicker().on('generated', function (element)
{
  $(element).append(...);
});```

then your event will be attached too late -- it'll already have fired by the time you attach a handler.

Attaching first might work:

```$('#datepicker').on('generated', function (element)
{
  $(element).append(...);
}).datepicker();```

... but that's a strange pattern and makes it easy for people to shoot themselves in foot.

What about a option like we already have for stuff like onBeforeShowDay? Could add an onGenerated option that would be a function called with the element. I think that'd be a bit more intuitive than the event.

Your thoughts? If that works, I'll rework this PR to match that model.

Azaret:
Well I'm not personally against to have to set `on` first, but I can understand people that want to follow the mainstream jQuery syntax ;) (Note that in JS best practices, a function should be declared before being used (which makes sense), with this rule in mind, events catcher should always be defined first).

What looks reasonable to me would be to have both for this particular event, as it as kinda an exception.
So we would have to test if we have the option `onGenerated` defined, if yes we call it, if no we trigger the event.
I think that doing both at the same time will do more damage than anything else, so calling one or the another seems like a good plan.

So for sure the documentation will need to be clear on that we cannot us both at the same time.
Also it should clearly state that for using the event instead of the option the user should define the event before instance the picker.

Issue #2037, opened 2016-10-14
Append div to specific dates


reachtokish:
Hi, Can I append div to some specific dates? if yes then how?

Thanks

Azaret:
Hi,

May I ask you your end goal ?
The easy way to 'catch' a date would be to add a specific CSS class with [`beforeShowDay`](https://bootstrap-datepicker.readthedocs.io/en/stable/options.html#beforeshowday).
Unfortunately there is no way push html code in `beforeShowDay`, nor an event triggered when dates change.
We might be able to help you with more information.

reachtokish:
Thanks for reply

I was trying to do a event calendar by bootstrap datepicker and I need a tooltip into some of the specific dates but Now I have choose another library for that.

Thanks :)

Azaret:
Yeah there is no build-in functionality for that.

A way I can think of would be to use `beforeShowDay` in order to tag the date you need to.
Then you could catch `DOMSubtreeModified` on the datepicker so you can inject your html code then.

It's a bit hacky but it might work well.

reachtokish:
Thanks a lot for your kind information :+1:

Azaret:
Brace yourself @reachtokish :)

acrobat:
Closing this issue as #2043 was merged. The finishing touches for this feature (other beforeShow methods) will be done in #2120

reachtokish:
ok

PR #2043, opened 2016-10-24
Allow customizing day cell value

programcsharp:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| License | MIT |


js/bootstrap-datepicker.js
```diff
@@ -1196,6 +1200,10 @@
 			}
 
 			if (!target.hasClass('disabled')){
+				// Allow clicking on elements inside a day
+				if (target.parent().hasClass('day'))
+			        target = target.parent();
```
Azaret: Since you allow html, you code won't work if I send `<div><div></div></div>`, you should change `e.target` by `e.currentTarget`, but then you should do a full test to be sure it does not break everything.

js/bootstrap-datepicker.js
```diff
@@ -1041,7 +1045,7 @@
 					clsName = $.unique(clsName);
 				}
 
-				html.push('<td class="'+clsName.join(' ')+'"' + (tooltip ? ' title="'+tooltip+'"' : '') + (this.o.dateCells ? ' data-date="'+(prevMonth.getTime().toString())+'"' : '') + '>'+prevMonth.getUTCDate() + '</td>');
+				html.push('<td class="'+clsName.join(' ')+'"' + (tooltip ? ' title="'+tooltip+'"' : '') + (this.o.dateCells ? ' data-date="'+(prevMonth.getTime().toString())+'"' : '') + '>' + displayDate + '</td>');
```
Azaret: You are actually making the other views sad and envious to the monthView :)
May you add this on the other views please ?

js/bootstrap-datepicker.js
```diff
@@ -1017,6 +1017,8 @@
 				clsName = this.getClassNames(prevMonth);
 				clsName.push('day');
 
+				var displayDate = prevMonth.getUTCDate();
```
Azaret: Since other tickets like #2037 would like such PR, I think a more abstract name for this option should be use, like `content`. Do you guys agree @acrobat @vsn4ik @hebbet ?

acrobat: :+1:

js/bootstrap-datepicker.js
programcsharp: I'm not sure that e.currentTarget would work because bubbling/propagation are stopped at the top of the click handler.

Instead, I use .parents(), does that work?

programcsharp:
* Changed option name to content
* Changed click handler to use .parents() to find .day
* Added generic test for content
* Added test for clicking inside nested content

Anything else need done here?

js/bootstrap-datepicker.js
```diff
@@ -1196,6 +1200,12 @@
 			}
 
 			if (!target.hasClass('disabled')){
+				// Allow clicking on elements inside a day
+				var day = target.parents('.day');
```
acrobat: Day is already defined on line 1183, see travis build

js/bootstrap-datepicker.js
Azaret: It should I guess

Azaret:
Yeah, I would like to see the same feature for the others views (month, year, decade, ...) lookup for `beforeShowMonth` for month and `callback` inside `_fill_yearsView()` for others.
Ty

programcsharp:
OK, I can do that for consistency and all that goodness. I'll fix the var name too.

Do you have a better suggestion for how to handle deeply nested clicks? I can duplicate the same thing I did for day -- ```target.parents(".day, .month, .year")```, but that just seems non-optimal.

Azaret:
Do it for the fame ;)

`target.closest()` should be better performance wise I think, since it will stop at the first parent match.
It should be enough for now, we might break the `click` function in the `2.x`, since binding events directly to the actual `td` will be easier to manage than catching any click anywhere on the picker (which by the way is a broken way of doing things because it does trigger the `show` event at any clicks).

bmb0205:
Being able to customize content within each day would be really awesome. I am trying to display counts of something in each day but haven't been able to figure out how.

acrobat:
@Azaret Can you review the pull request again? Thanks!

acrobat:
@programcsharp I would like to get this PR finished and have it shipped with 1.7.0! :rocket: Do you need to do any work here? Please rebase the branch against the current mast and can you check the failing travis build? Thanks!

@Azaret Time for a last review? Thanks!

programcsharp:
OK, lemme see what I can do here.

programcsharp:
Fixed the dupe variable name issue @Azaret mentioned. Should be ready to go now.

I don't have time to add the month/year/decade/century implementations.

js/bootstrap-datepicker.js
programcsharp: Fixed.

acrobat:
Thanks @programcsharp! I will look into this to see what I can do to have this feature shipped with 1.7.0

Azaret:
I guess we can push it in RC2 @acrobat

acrobat:
@Azaret Yes I will merge this today and tag a new RC release. But we need to try to implement this feature for all beforeXxx methods, I hope to find some time for this!

acrobat:
Thanks @programcsharp! The missing parts for beforeShowMonth (yeay, decade, etc) will be done in #2120

PR #2045, opened 2016-10-25
Removing sourceMappingURL from not-minified verson

### LINKED ISSUE #1795: remove sourceMappingURL form .css and .min.css builds perhaps?

esl-douglas:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #1795 
| License         | MIT

The not-minified version of the CSS doesn't need the source mapping. The mapping is not used when the full CSS version is being used.

Here where I work, we use this plugin with the not minified version, and Rails minifies and compile the CSS for us, so the source mapping on the full CSS doesn't make sense.

### DIFF
--- a/dist/css/bootstrap-datepicker.standalone.css
+++ b/dist/css/bootstrap-datepicker.standalone.css
@@ -507,4 +507,3 @@
 .datepicker.datepicker-inline td {
   padding: 4px 5px;
 }
-/*# sourceMappingURL=bootstrap-datepicker.standalone.css.map */


Azaret:
`/dist` is generated.
`Gruntfile.js` should be edited but I'm not sure it's a good idea at this point.

Azaret:
Maybe @vsn4ik can edit it, but I won't, those things break too easily.

PR #2054, opened 2016-11-07
adds failing test that setDatesDisabled resets viewDate to default one

jelhan:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | adds tests for #1980
| License         | MIT

Bug was reported in #1980

### DIFF
--- a/tests/suites/methods.js
+++ b/tests/suites/methods.js
@@ -144,6 +144,12 @@ test('setDatesDisabled', function(){
     this.dp.setDatesDisabled(['01-01-2011']);
     equal(monthShown.text(), 'March 2011', 'should not change viewDate');
     strictEqual(returnedObject, this.dp, "is chainable");
+
+    // upddate viewDate to a non default one, this is currently only possible with private api
+    this.dp._setDate(Date.UTC(2016, 0, 1), 'view');
+    equal(monthShown.text(), 'January 2016', 'testing assumption that _setDate with `view` as second parameter updates viewDate');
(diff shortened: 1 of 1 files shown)


acrobat:
Do you maybe have the time or an idea how to fix this bug? I won't merge this failing test as it work break our test suite, but with a fix (in this PR or in another/by another user) I would be happy to merge this!

jelhan:
I'm sorry but I don't have the time to fix this. It's not a low hanging fruit. I guess `update` method, which is used by several api methods, has to be refactored to address this issue. It's setting `[viewDate](https://github.com/uxsolutions/bootstrap-datepicker/blob/master/js/bootstrap-datepicker.js#L791-L800)` based on several conditions. Sometimes this is expected sometimes not.

Calling this methods use won't expect `viewDate` to change and we could consider them to be affected by this bug as well: `setStartDate`, `setEndDate`, `setDatesDisabled`, `setDaysOfWeekDisabled` and `setDaysOfWeekHighlighted`. Not quite sure if user expects viewDate to change on `clearDates`.

Note that this is not an issue if `updateViewDate` option is set to `false` (but `true` is default).

jelhan:
PR #1646 addresses this issue but wasn't merged.

acrobat:
Thanks @jelhan for the heads up! I've requested a rebase and review in the PR. So hopefully these 2 pr's can be merged!

PR #2063, opened 2016-11-12
Add event dispatch to trigget eventChangeHandler in React

### LINKED ISSUE #1152: M month format
Looks like it doesn't allow a month format of M (i.e. Jan Feb etc.)

This is important for international sites as 01/05/14 means either 1st of May or 5th of Jan

murilomothsin:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #1152
| License         | MIT

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -601,6 +601,8 @@
 		setValue: function(){
 			var formatted = this.getFormattedDate();
 			this.inputField.val(formatted);
+			var event = new Event('input', { bubbles: true });
+			this.inputField[0].dispatchEvent(event);
 			return this;
 		},
 


acrobat:
Anyone with react experience to review this pr?

\cc @vsn4ik @Azaret @hebbet @jelhan

jelhan:
I'm not a react guy but this PR would not only affect react. It would trigger an `input` event if value changes. This could be a breaking change since an application might assume that `input` event is only triggered if user changes data without using using bootstrap-datepicker UI (entering it directly in input field).

@murilomothsin could you please elaborate why why react needs an `input` event and isn't satisfied by `change` event?

If we decide to trigger `input` events on every change they should be triggered at same place as `change` events. Otherwise code would be hard to maintain. Also PR is missing any tests.

Azaret:
Not a react guy either. And when googling eventChangeHandler I come back here 😞  
Also the issue is not related what so ever.  
Then tests are not validating.  
So this PR is getting close in my eyes.  

That being said, updating Events dispatch according to latest RFC's is something we should take into consideration.  
And in this specific case, we should not dispatch the `change` event, but `input`.
Since `change` should be dispatched when the user is committing a value.

PR #2065, opened 2016-11-16
Fix the visibility of the prev and next arrows for years views

### LINKED ISSUE #2058: ViewMode 2: can access the previous/next decade where all years are disabled
### Expected behaviour
When the startView is set to 2, the upper arrows pointing to the previous or next decade should not be visible if that all the years in that decade are disabled due to the startDate/endDate options.

### Actual behaviour
Unless `selectedYear == startDate.year`, the user can click on the upper left arrow and go to the previous decade where all options are disabled.
Similary, unless `selectedYear == endDate.year` the user can click on the upper right arrow and go to the next decade where all options are disabled.

### Datepicker version used
1.6.4

### Example code
https://jsfiddle.net/ua48nx09/4/

In this example you can go back to the previous decade and forward to the next decade, although that should be disabled. 

If you select a date with the `year = startDate.year` (2014), the previous decade will be disabled as it should be. Similarly if you choose a date with `year = endDate.year` (2016), the next decade will be disabled.

### Location of the problem in the code

The problem seems to be in the `updateNavArrows` function and especially here:
```javascript
if (this.o.startDate !== -Infinity && year <= (this.o.startDate.getUTCFullYear()) || this.o.maxViewMode < 2) {
    this.picker.find('.prev').css({visibility: 'hidden'});
}
```
and here:
```javascript
if (this.o.endDate !== Infinity && year >= this.o.endDate.getUTCFullYear() || this.o.maxViewMode < 2){
    this.picker.find('.next').css({visibility: 'hidden'});
}
```

Azaret:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #2058

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1136,8 +1136,8 @@
 			var d = new Date(this.viewDate),
 				year = d.getUTCFullYear(),
 				month = d.getUTCMonth(),
-				prevState,
-				nextState;
+				prevState, nextState,
+				factor = 1;
 			switch (this.viewMode){
 				case 0:
 					prevState = (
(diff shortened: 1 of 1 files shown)


js/bootstrap-datepicker.js
```diff
@@ -1152,18 +1152,24 @@
...
+					factor *= 10;
+					/* falls through */
+				case 2:
+					factor *= 10;
+					/* falls through */
+				case 1:
 					prevState = (
 						this.o.startDate !== -Infinity &&
-						year <= this.o.startDate.getUTCFullYear()
+						(parseInt(year / factor, 10) * factor) <= this.o.startDate.getUTCFullYear()
```
vsn4ik: `Number` faster.

vsn4ik: `(pa... ...tor)` brackets are not needed.

Azaret: Is it really ? Good to know.

vsn4ik: https://jsperf.com/math-floor-vs-math-round-vs-parseint/181

vsn4ik:
Thanks @Azaret. Tests added in #2101.

Issue #1550, opened 2015-09-01
Use calculated dates for defaultViewDate


stantond:
`defaultViewDate` only works with an actual date, for example, `{'year': '1990'}`. Other settings can be defined as `-25y` for the same result. Can this functionality be added to `defaultViewDate` please?

Thanks

vsn4ik:
Use `defaultViewDate: '-25y'` on next release. Thanks @stantond!

PR #2070, opened 2016-11-19
Allow Date/String for defaultViewDate option

### LINKED ISSUE #1611: Option defaultViewDate as data attribute
Hi folks,

can somebody tell me how to set the defaultViewDate as an data attribute?

I tried the following but did not work:

```
       data-date-default-view-date-year="1977" 
       data-date-default-view-date-month="0" 
       data-date-default-view-date-day="1"
```

CherryDT:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | fixes #1611 (at least for the OP's use case where all three values are set) and #1550
| License         | MIT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,6 @@
 Changelog
 =========
 
-1.7.0
------
-
-Features
- * Added support for dates/strings to defaultViewDate option
-
 1.5.1
 -----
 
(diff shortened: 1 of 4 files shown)


Azaret:
Looks good but can you revert the changelog file please ?
It is actually not used anymore since 1.5.1 and for the moment we need to not add confusion about our changelog 'management'.

CherryDT:
There you go

vsn4ik:
Thanks @CherryDT!

PR #2095, opened 2016-12-24
Add Code Of Conduct

ivannieto:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,74 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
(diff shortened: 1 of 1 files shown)


acrobat:
@hebbet @vsn4ik @Azaret What do you guys think? I think it's a good idea to have this in the repo, it feels natural to me that this should be the guidelines but it does no harm to have them explicitly added

ivannieto:
@acrobat thanks for your point of view. I think it's really the intention of having this file separately so README's and CONTRIBUTING files remain clean. It's not intention of this file to say how to behave as humans, but good guidelines on how to feel part of the community. 

Also, it can be edited as the project needs.

PR #2098, opened 2017-01-04
Some improvements

vsn4ik:
1. Remove unneeded class


js/bootstrap-datepicker.js
```diff
@@ -1820,11 +1821,11 @@
 					fn = fn_map[part[2].toLowerCase()];
 					date = Datepicker.prototype[fn](date, dir);
 				}
+				// Use Datepicker.prototype._zero_utc_time in future
```
acrobat: Is this just some extra info or something we should do in the future? If so add `TODO:` before the comment so we can find those todo's easily

less/datepicker.less
```diff
@@ -5,7 +5,7 @@
 		width: 220px;
 	}
 	direction: ltr;
-	&&-rtl {
+	&-rtl {
```
vsn4ik: Before:
```
.datepicker.datepicker-rtl {
  direction: rtl;
}
.datepicker.datepicker-rtl.dropdown-menu {
  left: auto;
}
.datepicker.datepicker-rtl table tr td span {
  float: right;
}
```

After:
```
.datepicker-rtl {
  direction: rtl;
}
.datepicker-rtl.dropdown-menu {
  left: auto;
}
.datepicker-rtl table tr td span {
  float: right;
}
```

Tested on demo page with `ar` language - work fine.

acrobat: Isn't this a somewhat soft BC break as users could use the `.datepicker.datepicker-rtl` selector. I also think this a bug but by fixing it we break BC. What do you think? \cc @Azaret @hebbet

Azaret: Don't think much about it. Tho I'm not really sure to know how it can BC.   
jQuery/JS class selectors will still work since the classes will still be here.  
If someone has a CSS rule override ours, it will still works since this new rules has less weight (010/020/010) than the one before (020/030/020).

PR #2114, opened 2017-02-02
added clearDates() so .data('datepicker').clearDates() clears range

coderofsalvation:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | https://github.com/uxsolutions/bootstrap-datepicker/issues/1724
| License         | MIT


acrobat:
Hi @coderofsalvation, thanks for the PR. Can you check the failing travis build for codestyle issues.

coderofsalvation:
done

js/bootstrap-datepicker.js
```diff
@@ -1543,6 +1543,10 @@
 				p.setRange(range);
 			});
 		},
+		clearDates: function(){
+				this.pickers[0].clearDates();
+				this.pickers[1].clearDates();
```
acrobat: I'm not sure if we should check for the 0 and 1 key or just loop over all `this.pickers`. In theory there should only be 2 pickers, what do you think @coderofsalvation 

/cc @Azaret @vsn4ik

coderofsalvation: dateRange *always* has two pickers, so the assumption of 0 and 1 are safe.
If it would be more, it wouldn't be a daterange.

acrobat:
@coderofsalvation Can you rebase the PR against the current master? Thanks!

js/bootstrap-datepicker.js
```diff
@@ -1545,6 +1545,10 @@
...
```
KrunchMuffin: I believe this to be an issue, but maybe it isn't. When calling this `changeDate` also gets called 6 times along with another `clearDate`.  To me the date isn't changing per se, it's being removed.
image

PR #2122, opened 2017-02-14
Fix beforeShowDay with content not equal number

vsn4ik:
After #2043.
Use data-date attribute always.


Azaret:
Took me a while to get what you did, but that's actually way more clever and safe.
Kudos.

PR #2123, opened 2017-02-14
Refactor _setDate params in dayCellClick

vsn4ik:
After #2122.


Azaret:
Yeah the click function will need some serious rethinking with all those editable templates.
Thank you @vsn4ik !

PR #2124, opened 2017-02-14
WIP: Expand customizable cells to all views

Azaret:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | rather enhancement
| BC breaks?      | no
| Related tickets | fixes #2043, partially #2043
| License         | MIT

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -49,6 +49,7 @@ A function that takes a date as a parameter and returns one of the following val
    * ``enabled``: same as the Boolean value above
    * ``classes``: same as the String value above
    * ``tooltip``: a tooltip to apply to this date, via the ``title`` HTML attribute
+   * ``content``: the content to display in the month cell, rather than the default (month label as text)
 
 
 beforeShowYear
@@ -66,6 +67,7 @@ A function that takes a date as a parameter and returns one of the following val
    * ``enabled``: same as the Boolean value above
    * ``classes``: same as the String value above
    * ``tooltip``: a tooltip to apply to this year, via the ``title`` HTML attribute
+   * ``content``: the content to display in the year cell, rather than the default (year as text)
 
 
 beforeShowDecade
@@ -83,6 +85,7 @@ A function that takes a date as a parameter and returns one of the following val
    * ``enabled``: same as the Boolean value above
    * ``classes``: same as the String value above
    * ``tooltip``: a tooltip to apply to this year, via the ``title`` HTML attribute
+   * ``content``: the content to display in the decade cell, rather than the default (decade as text)
 
 
 beforeShowCentury
@@ -100,6 +103,7 @@ A function that takes a date as a parameter and returns one of the following val
    * ``enabled``: same as the Boolean value above
    * ``classes``: same as the String value above
    * ``tooltip``: a tooltip to apply to this year, via the ``title`` HTML attribute
+   * ``content``: the content to display in the century cell, rather than the default (century as text)
 
 
 calendarWeeks


Azaret:
@vsn4ik I took the liberty to enhance your idea.  
I directly attach the event listener to the correct DOM node and use jQuery data() so the DOM is kept clear.

Azaret:
Attaching the event directly to the node was in my mind for the v2.
But the click function is already less a mess than 2 minors versions ago, so it's safer to make the change now.

Azaret:
@acrobat Fill free to play around to be sure I didn't break everything.

Some snippet for your amusement:
```javascript
function callback() {
    var smileys = ['😁', '😋', '😡', '😰', '🙅', '🙊', '🙏', '✌', '🚀', '🚨', '🚧', '🚽', '🉐', '⌚', '☝', '⚽', '⚾', '⛄', '⛅'];
    return {content: '<div>' + smileys[Math.floor(Math.random() * smileys.length)] + '</div>'};
}

$('input').datepicker({
    beforeShowDay:     callback,
    beforeShowMonth:   callback,
    beforeShowYear:    callback,
    beforeShowDecade:  callback,
    beforeShowCentury: callback
});
```

acrobat:
@Azaret Thanks for the work! I've tested the code and it looks like it works for all views (cool example btw :D)

@vsn4ik @hebbet Do you guys have time for a review?

acrobat:
@Azaret Do we need extra tests for this behavior (the changes to the click etc) or is it covered by the current tests?

js/bootstrap-datepicker.js
```diff
@@ -842,14 +842,17 @@
...
+			var focused, node, localDate = this._utc_to_local(this.viewDate);
+			this.picker.find('.datepicker-months td').html('');
 			for (var i = 0; i < 12; i++){
 				focused = localDate && localDate.getMonth() === i ? ' focused' : '';
-				html += '<span class="month' + focused + '">' + dates[this.o.language].monthsShort[i] + '</span>';
+				node = $('<span>');
+				node.data('date', {day: 1, month: i, year: this.viewDate.getUTCFullYear()});
+				node.attr('class', 'month' + focused);
+				node.html(dates[this.o.language].monthsShort[i]);
+				node.on('click', $.proxy(this.cellClick, this));
```
vsn4ik: No, it is bad practice.

Azaret: Why ?

vsn4ik: 90 listeners (your option) vs 4 listeners.

acrobat:
@Azaret @vsn4ik I'm going to release RC2 in the next few days. If this is complete then it will be in RC otherwise we will do another RC when this is merged!

Azaret:
Well it looks clear to me, I'm waiting @vsn4ik input on his remarks.

acrobat:
Hmm he had posted a comment, but it looks like it was deleted. So indeed wait for @vsn4ik for his feedback

js/bootstrap-datepicker.js
vsn4ik: If you follow your rule will turn 90 listeners: `(4 * 12) + 42 (days) = 90`. Is not critical, it is possible to leave.

acrobat: Is this comment fixed?

Azaret: Yeah I though about it the other day and it hit me what @vsn4ik was talking about.  
The issue is that by design it is not easy to catch the event when the DOM is customized.  
Comprehensive drawing:  
image

Before it was easy, you can bind on the container, and get the source of the event with `currentTarget`.  
Now `currentTarget` will return the custom DOM nodes instead of the td cell.

The easiest and safest way is then to bind the event to the cell, so `target` will equals the cell, always.    
If we keep the event bound to the container, it mean we'll have to lookup for the cell between the  `target` and the `currentTarget`.  

So the question is, which is the best performance wise between binding 90 events or recursively find  a DOM element.

acrobat: Ok that explains a lot! So we should go for the 90 events for now? What do you think @vsn4ik?

bmb0205:
When is this expected to come out? According to the docs, returning an object with a content key in the beforeShowDay method should allow me to change the content, however it does not work. I can disable a day fine with a boolean, just can't set content.

acrobat:
@Azaret sorry to ping you again! But any more work needed here? If I understand your comment above correctly "90 events" choice we have to make is good enough for now?

If we can agree on some solution for this would be nice, this way I can put out a RC2 ( and hopefully quite quick a stable release)

/cc @vsn4ik

js/bootstrap-datepicker.js
acrobat: Ping @vsn4ik

PR #2130, opened 2017-02-20
Fix rtl position problem

### LINKED ISSUE #2127: position proble in component mode for fa language and top right orientation
### Expected behaviour
show datepicker in top right of component button

### Actual behaviour
show datepicker in an incorrect position in left side of input

### Datepicker version used

v1.6.1

### Example code
https://jsfiddle.net/r5x4enym/

imansa:
to fix picker position in rtl mode for rtl page
issue#2127
https://github.com/uxsolutions/bootstrap-datepicker/issues/2127

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #2127
| License         | MIT


acrobat:
What do you guys think of this change @vsn4ik @Azaret? I don't know it for sure as we had many changes to this part of the code to fix various issues (but sometimes broke other cases)

vsn4ik:
Wrong branch. Need `master`.

acrobat:
That's something I can change, let's see if this works without conflicts

Azaret:
This issue is for `1.6.1`.  
It seems to work fine on `master`.  

Are we going to do LTS on 1.6 ?

acrobat:
hmm I thought the issues was also present in master :s I'm not in favor of an LTS on 1.6, if we really should do an lts it should be on 1.7 (if this is the last minor release before 2.0)

In this case I would say upgrading to 1.7 should fix this issue (stable release should be out soon, only few last pr's). If this is not possible we should push an extra patch release for 1.6

Azaret:
Neither I am. Tho 1.7 is not 100% ready, it is decently usable, I'm in favor pushing people into 1.7 as we fixed many issues that the 1.6 branch had.

acrobat:
So an extra path release for 1.6 is what we should do here? I have no problem with that btw!

Azaret:
No, I just tested it against `1.6.4` and it work fine too:
image

acrobat:
@imansa can you provide some feedback as it looks like it works on 1.6.4?

alexandereneroth:
This issue is still present in 1.7

Issue #2148, opened 2017-03-20
setDefaultViewDate could be nice


hersing:
### Expected behaviour
It would be nice if there was a method called _setDefaultViewDate_, similar to _setStartDate_, that made it possible to change the _defaultViewDate_ option.

I have created my own period/range calendar, and i want to change the _to date_ datepickers _defaultViewDate_ to the value of the the selected _from date_

### Datepicker version used
1.6.4.

###  code
If I breakpoint in the _setStartDate_ and run `this._process_options({startDate: startDate});` in the console it seams to work. So can this be implemented easily?

hersing:
Or even better, create a _setOptions_. this will then run the code `this._process_options(newOptions);`

hersing:
I just made a Pull request on this.

PR #2149, opened 2017-03-21
new Method: setDefaultViewDate add

hersing:
| Q                    | A
| --------------- | ---
| Bug fix?          | no
| New feature? | yes
| BC breaks?     | no

I added a _Method_ called _setDefaultViewDate_, that changes the defaultViewDate.


acrobat:
@hersing Thanks for the PR! Can you add a test to cover this new behavior?

hersing:
hey @acrobat,
I already added a test in the _\tests\suites\methods.js_ file on line 95. Do you want more tests then that?

acrobat:
Oh sorry I've overlooked the check in the tests. I thought it was only a test to see if it is chainable!

jelhan:
I'm not convinced about the approach. As far as I understand `defaultViewDate` should only be used on initialization. It shouldn't have any impact on already initialized datepicker. @hersing isn't this more about a missing public api to change `viewDate` as suggested in #1978?

Also as far as I understand current behaviour is not like described in #2148. In range mode bootstrap-datepicker sets date selected in the other datepicker if otherwise it wouldn't be a range anymore (there wasn't any date selected before, start time is after end time, or end time is before start time). If bootstrap-datepicker updates date selected `viewDate` is updated accordingly to newly selected date as long as `updateViewDate` is not `false` (which isn't the default).

All in all I didn't see a use case for this api method. Okay, perhaps this one could be used as a work-a-round for #2054 but a fix to this bug would be better. ;-)

hersing:
I think you might have misunderstood my use for it.
I made the method to be able to change the default view in runtime, so I don't have to run initialization again and again. At my company the datepickers  default view has to change depending on other choices made on the screen...

Let's say we have a page saying "choose your star sign" and according to the chosen sign the date picker should have a specific default month to start in. I would find it problematic that you have to initialize the date picker every time a new sign is chosen in the "star sign select".. (this is just a bad example).

Do to a lot of company rules and the possibility for mistakes I have to use to date pickers to make our range, and not use the datepickers range mode.

Since the method I made seems to work flawlessly I don't see the big problems.

jelhan:
@hersing For me that sounds like an `setViewDate` method. Also your code looks like you are trying to implement a solution like that one cause it not only changes `defaultViewDate` but also runs an `update`. Since `update` changes `viewDate` (which is indeed causing several bugs, #2054) this might actually work in some cases but is not what I would expect to happen if I change the `defaultViewDate`. Indeed it would reset `viewDate` to last selected date or if no date is selected yet to `defaultViewDate`. (There might also be edge cases where it resets `viewDate` depending on `startDate` or `endDate` but I'm not quite sure if this could actually happen.)

Could you please provide a JSFiddle demonstrating your use case using internal api? Also it would be nice if your test not only covers that `defaultViewDate` option is changed internally but also testing how rendered datepicker is affected.

hersing:
Well a setViewDate seams like a better idea. All I want to do is beeimg able to change the page you see when opening the datepicker without have to set the date first. In my company a empty date means something to :) 

I can try to look at a JSFiddle when I got the time. :) 
I'm working on some governmental stuff with a deadline on July 1th. so we are a bit busy.

hersing:
This isen't pretty at all, and it's just a fast JSFiddle... But as I said I can't use your range, but I have to have a range datepicker, so my example just quickly shows how i won't the second date picker to change the viewdate when the date on the other one changes

[https://jsfiddle.net/bz1xxff8/1/](https://jsfiddle.net/bz1xxff8/1/)

jelhan:
I'm still not quite sure what you are trying to achieve but I modified your JSFiddle to show how you could do it by using private api: https://jsfiddle.net/j09jr7ob/1/

#1978 does not suggest more than exposing `.datepicker('_setDate', date, 'view')` as public api method `setViewDate`.

hersing:
Could work.. I will try it out....
One small bug though! When selecting a date, the "to date" starts on the month before... is that a bug?

jelhan:
Yeah, you are right. `e.date` is locale date while `.datepicker('_setDate', date, 'view')` expects an UTC date. So you have to convert e.date to UTC first.

nikonov91-dev:
Hi, was this feature released?  thanks

avsdev-cw:
Bit necro-ish but Just diving in on this from the world of shiny: having a setOption or setDefaultViewDate is extremely useful since most UI code is defined on service start. Like the OP we have a requirement to change the initial view date depending on a user selection (in this case an entry from a table row).

PR #2154, opened 2017-03-24
Added new option: multidateAutoRemove

flamewave:
Added a new option where when using the multidate option as a number, you can enable/disable the automatic dropping of the oldest selected dates. The option defaults to true so that unless the option is explicitly set, the current functionality (without the option) is used.
New option: multidateAutoRemove
Type: Boolean
Default value: true

| Q               | A
| --------------- | ---
| Bug fix?        |no
| New feature?    |yes

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -1279,11 +1279,11 @@
 				this.dates.clear();
 				this.dates.push(date);
 			}
-			else {
+			else if (this.o.multidate === true || (typeof this.o.multidate === 'number' && (this.dates.length < this.o.multidate || this.o.multidateAutoRemove))) {
 				this.dates.push(date);
 			}
 
-			if (typeof this.o.multidate === 'number')
+			if (this.o.multidateAutoRemove && typeof this.o.multidate === 'number')
 				while (this.dates.length > this.o.multidate)
 					this.dates.remove(0);
 		},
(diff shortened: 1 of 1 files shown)


vsn4ik:
`multidate: Number`?

flamewave:
Yes, it's used in conjunction with `multidate: Number`

Example:
`$("#my-input").datepicker({
  multidate: 5,
  multidateAutoRemove: false
});`

jelhan:
@flamewave Could you please provide a use case? Why should someone use `{ multidate: 5, multidateAutoRemove: false }` and not `{ multidate: true }`?

flamewave:
When you use `{ multidate: true }`, it allows you to select an unlimited number of dates.

When you use `{ multidate: 5 }`, it allows you to select up to 5 dates, and if you select a 6th date, it unselects the first date you had selected and selects the new date instead. As the docs say: 

>  If a number is given, the picker will limit how many dates can be selected to that number, dropping the oldest dates from the list when the number is exceeded.

When you use `{ multidate: 5, multidateAutoRemove: false }`, it allows you to select up to 5 dates, and if you try to select a 6th date it will not allow you. In other words, it disables the "dropping the oldest dates from the list when the number is exceeded" functionality.

Let me know if you need anything further.

jelhan:
That sounds reasonable. Would be nice if you also update documentation and add tests in your PR.

flamewave:
Docs updated, might be awhile before I get to the tests, I've been quite busy lately.

PR #2177, opened 2017-04-27
Update markup.rst

edwin-wei:
I found that in current version, 'input-daterange' actually implements the datepicker method, and the code example below runs well.
`$('.input-daterange').datepicker()` 
On the other hand, the original code example proposed, i.e. 
`$('.input-daterange input').each(function() {
    $(this).datepicker('clearDates');
});` , 
breaks the connection between "start" and "end" inputs. For example, it does not add proper listener on "changeDate" event.

| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | no|yes
| BC breaks?      | no|yes
| Related tickets | fixes #X, partially #Y, mentioned in #Z
| License         | MIT

### DIFF
--- a/docs/markup.rst
+++ b/docs/markup.rst
@@ -51,13 +51,11 @@ Using the ``input-daterange`` construct with multiple child inputs will instanti
 .. figure:: _static/screenshots/markup_daterange.png
     :align: center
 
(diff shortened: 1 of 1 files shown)


acrobat:
I'm not sure about this change. See [example](http://jsfiddle.net/zcvq0yhh/284/). If it should work, can you provide a jsfiddle to show it? Thanks!

PR #2205, opened 2017-06-15
blur inputField on hide so focus event can fire again

mreiden:
The datepicker is hidden if you scroll while it is open, but the inputField remains focused.  The focus prevents it from showing again if you click on the input because the onFocus event does not fire until you click off of the input and then click it again.

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -502,6 +502,7 @@
 
 			if (this.o.forceParse && this.inputField.val())
 				this.setValue();
+			this.inputField.blur();
 			this._trigger('hide');
 			return this;
 		},


acrobat:
@mreiden thanks for the PR! Can you provide a jsfiddle to show the problem? Thanks!

mreiden:
To see the problem, open the datepicker and then click the scrollbar.

[1.7.0-RC3 with the problem](https://jsfiddle.net/bdsmdytg/1/)

[with the blur on hide](https://jsfiddle.net/bdsmdytg/2/)

acrobat:
@mreiden thanks for the examples! And what actions do I need to do to reproduce the behavior?

mreiden:
1) Click to open date picker.

2) Click on or scroll the document.  This should close the date picker but leaves the input focused (FireFox, Chrome, and iOS safari).  

3) Try clicking to open the date picker again.  This fails to open in FireFox, Chrome, and Safari.  It opens as expected in IE.

The blur on hide link should work as expected.

PR #2243, opened 2017-08-05
Use $.fn.css instead of $.fn.(show|hide)

stash86:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

I'm using jQuery 3.x, and the popup windows just won't show. I looked around google, and found that the solution is just by changing .show() into .css('display','block'). Now it's working perfectly with jQuery 3.x. I have tested it with jQuery 1.12.4 and jQuery 2.2.4, and it's still working normally.


vsn4ik:
@stash86 thanks for the PR!
Can you provide a jsfiddle to show the problem?
Thanks!

stash86:
Sure, I'll try to create a jsFiddle to show the problem

stash86:
Seems like my problem is a localized problem on my production site only. My localhost site and my jsfiddle shows no error using the original js file. You can delete this PR if you want.

vsn4ik:
Thanks @stash86!

PR #2247, opened 2017-08-16
Add setting to allow explicit setting of isInline

lukeburden:
I recently upgraded to 1.7.1 and discovered `autoclose` was no longer being respected.

Our use case involved a div as a trigger surrounding an input field that we populate on a `changeDate` event. Due `datepicker` being called on that div rather than an input, the `isInline` logic was resolving to `true`, which then essentially disables the `autoclose` behaviour.

I'd like to suggest a setting to force `isInline` such that developers can better control this behaviour.


lukeburden:
@vsn4ik any thoughts on this PR?

vsn4ik:
Thanks @lukeburden!

PR #2367, opened 2018-04-10
remove always false condition from changeDate when manually typing in input

McNetic:
This is a regression since version 1.6.1. At least in 1.7.1 and 1.8.0, the changeDate event is broken when manually entering/changing the date in the input field. The problem is the if ... else if:
```
if (fromArgs) {
} else if (....) {
  if (... && fromArgs) {
  }
}
```
Obviously, the second branch can never be taken. Maybe there should be an additional condition, but I don't see why the second condition contains fromArgs.

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -102,6 +102,7 @@
 
 	var Datepicker = function(element, options){
 		$.data(element, 'datepicker', this);
+		this._initialized = false;
 		this._process_options(options);
 
 		this.dates = new DateArray();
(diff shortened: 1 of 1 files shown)


McNetic:
Could I get some feedback on this PR? Meanwhile, I fixed the issue proper and also added tests ...

diesl:
@acrobat, @vsn4ik could you please check?

PR #2399, opened 2018-06-29
Move JQuery to peerDependencies

isBatak:


CherryKitty:
Please fix

Azaret:
jQuery is not a peer dependency, why would you like it to be changed ?

PR #2444, opened 2018-11-05
Update docs and scheme

vsn4ik:


acrobat:
Looks good to me. Only remark in the file `docs/_static/screenshots/option_language.png` it seems that the chinese language is not working anymore. Possibly not related to these changes, but a possible bug

vsn4ik:
@acrobat Thanks! Fixed.

damu9618:
Hi @vsn4ik We are this date picker in one of the application. now we need to change the date prev and date next icons from 
image
 TO 
image
 something like this, can you please let me know whether is it possible or not

PR #2482, opened 2019-04-29
Correct issues with date range

### LINKED ISSUE #2473: Datepicker issue with Date Range
Hi

In the below scenario Date Range is not working correctly for Bootstrap Datepicker:
Step 1: Select Start Date E.g. 04/01/2019 so by default datepicker set 04/01/2019 date as the End Date.
Step 2: Now Clear Start Date using Clear Button. E.g Clear 04/01/2019 from Start Date.
Step 3: No select different Start Date which is greater than End Date E.g.  Select 04/05/2019 as Start Date. So End should get updated to 04/05/2019. But it is not working. If select 04/05/2019 as Start Date again then it is working.

See Below Video:
https://drive.google.com/file/d/1pPfQhwCxoZLeaDG_zAy8l6qhQr1JGMSZ/view

Thank You

### LINKED ISSUE #2481: Range - decreasing upper bound replaces lower bound even if lower bound was previously empty
### Expected behaviour
The lower bound should not be replaced with the upper bound when the lower bound was empty and the upper bound is being decreased.

### Actual behaviour
The empty lower bound is replaced by the upper bound.

### Datepicker version used
1.8.0

### Example code
https://jsfiddle.net/johnnyRose/okewcnpm/21/

Click the "Problem 1" button to reproduce. The fiddle above also reproduces #2473 by clicking the "Problem 2" button.

johnnyRose:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #2473, fixes #2481
| License         | MIT

https://jsfiddle.net/johnnyRose/okewcnpm/21/


vsn4ik:
Thanks @johnnyRose!

PR #2491, opened 2019-06-03
fix: allow any HTML element as inline container

ocarreterom:
| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | yes
| BC breaks?      | no|yes
| Related tickets |
| License         | MIT

Allows any element to be a container of inline date picker, even a custom element.

```html
<span data-date="2019-06-01"></span>
<section data-date="2019-06-01"></section>
<my-custom-element data-date="2019-06-01"></my-custom-element>
```


vsn4ik:
Thanks @ocarreterom!

PR #2498, opened 2019-07-19
Update Puppeteer

vsn4ik:


vsn4ik:
@acrobat Good catch! Fixed 🙂

acrobat:
Good job @vsn4ik! 

Just out of curiosity previously we used phantom-js, what is the advantage of puppeteer?

vsn4ik:
https://phantomjs.org/ 
> PhantomJS development is suspended https://github.com/ariya/phantomjs/issues/15344

acrobat:
Good to know! Btw thanks for your efforts on this library!

PR #2500, opened 2019-07-22
Added large initial captital

OscarKolsrud:
| Q               | A
| --------------- | ---
| Bug fix?        no
| New feature?    no
| BC breaks?      no
| Related tickets 
| License         | MIT


OscarKolsrud:
This only affects Norwegian

Azaret:
@Nettsentrisk @OscarKolsrud Is there any official references saying that month and weekdays should be capitalized or not ?

OscarKolsrud:
@Azaret No, there are not. But since they stand alone it makes so much more sense to have them capitalized than not. To make a difference between them, it looks so cluttered with everything in small letters. I prefer it capitalized rather than following the baselines

Nettsentrisk:
This PR should be rejected, as Norwegian language rules dictate that these should be as they are currently implemented. Here is an explainer from the official Norwegian language council: https://www.sprakradet.no/sprakhjelp/Skriveregler/Stor_eller_liten_forbokstav/#dagar

PR #2504, opened 2019-07-27
Add snippets section to readme file

Azaret:
Add a snippets section in README file with as first snippet the classic booking demo with two date picker

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | many
| License         | MIT


acrobat:
I like the idea! Maybe we should create a separate docs page where we can list all handy snippets. This way we can link to it from the readme and the are available through the online docs section. What do you think?

acrobat:
I'm going to merge this as is, we can add it later to the docs. Thanks @Azaret!

PR #2547, opened 2020-03-26
todayHighlight defaults table entry wrong

dataxpress:
todayHighlight is actually `false` by default; the defaults table incorrectly showed it as being `true` in conflict with the documentation higher up on the page.

| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT


vsn4ik:
Thanks @dataxpress!

PR #2549, opened 2020-03-26
Move hide call out of keydown tab event, into blur event

### LINKED ISSUE #2548: Mobile Next button doesn't fire keydown event, calendar stays open
### Expected behaviour
When on a mobile device, hitting the next button to focus on the next input should hide the calendar popup.

### Actual behaviour
When hitting the next button, focus is shifted to the next input in the form, but the calendar popup stays open until you trigger the mousedown event to call the hide() function.

### Datepicker version used
1.9.0

### Example code
Examples are hard to share as you have to open them on a mobile device or emulator. Just any date input with another text input after it in a form. The mobile keyboards enter key will become the next key when focused on the date input.

We can solve this by moving the call to hide() out of the keydown tab event, and into the inputs blur event. It will still get called when tabbing just in the subsequent event, and will then work with mobile next buttons.

jblakley89:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #2548 
| License         | MIT

### DIFF
--- a/js/bootstrap-datepicker.js
+++ b/js/bootstrap-datepicker.js
@@ -405,6 +405,7 @@
 				[this.element, {
 					blur: $.proxy(function(e){
 						this._focused_from = e.target;
+            this.hide();
 					}, this)
 				}]
 			);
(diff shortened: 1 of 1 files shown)


jblakley89:
As I've stated in the linked issue, this PR doesn't solve the problem and introduces another. There is a workaround in the issue thread.

PR #2647, opened 2022-01-17
Fix for Bootstrap 5.x

dannytoma:
Added input-group-text class to get this working with Bootstrap 5.x


hebbet (approved):

fralik:
For those coming here for Bootstrap 5 support. One way of doing it without library code modifications is to do like this:
```html
<div class="input-group date" data-provide="datepicker">
  <input type="text" class="form-control">
  <span class="input-group-text input-group-addon">
    <i class="bi bi-calendar3"></i>
  </span>
</div>
```

PR #2653, opened 2022-03-16
Fix translation of month of January for Ukrainian localization

YegorPo:
| Q               | A
| --------------- | ---
| Bug fix        | yes


tkoetkoetkoee:
datepicker.min

